PR #41, opened 2023-05-25
fix: multicall tests draft

0xmikko:
- small changes in CF


contracts/test/helpers/CreditFacadeTestHelper.sol
```diff
@@ -132,14 +132,14 @@ contract CreditFacadeTestHelper is TestHelper {
         vm.roll(block.number + 1);
     }
 
-    function multicallBuilder() internal pure returns (MultiCall[] memory calls) {}
+    function MultiCallBuilder() internal pure returns (MultiCall[] memory calls) {}
```
lekhovitsky: is it by intent or a result of automatic renaming?

contracts/interfaces/ICreditFacadeMulticall.sol
```diff
@@ -28,7 +28,7 @@ uint256 constant INCREASE_DEBT_WAS_CALLED = 2 ** 193;
 uint256 constant EXTERNAL_CONTRACT_WAS_CALLED = 2 ** 194;
 
 // Pay bot permisson is set there because it's used for bots only
```
lekhovitsky: probably worth updating the comment

lekhovitsky (approved):

contracts/test/helpers/CreditFacadeTestHelper.sol
0xmikko: Fixed

contracts/interfaces/ICreditFacadeMulticall.sol
0xmikko: fixed

PR #44, opened 2023-05-26
feat: pool fixes, optimization, refactoring and docs

lekhovitsky:


contracts/interfaces/IPoolV3.sol
```diff
@@ -1,127 +1,259 @@
...
-/// More: https://dev.gearbox.fi/developers/pool/abstractpoolservice
-interface IPoolV3 is IPoolV3Events, IPoolBase, IERC4626 {
+    /// @notice Deposits given amount of underlying tokens to the pool in exchange for pool shares
+    /// @param assets Amount of underlying to deposit
+    /// @param receiver Account to mint pool shares to
+    /// @return shares Number of shares minted
+    function deposit(uint256 assets, address receiver) external override returns (uint256 shares);
+
+    /// @dev Same as `deposit`, but allows to specify the referral code
     function depositReferral(uint256 assets, address receiver, uint16 referralCode) external returns (uint256 shares);
```
0xmikko: depositWithReferral?

contracts/interfaces/IPoolV3.sol
```diff
@@ -1,127 +1,259 @@
...
+    /// @notice Redeems given number of pool shares for underlying tokens
+    /// @param shares Number of pool shares to redeem
+    /// @param receiver Account to send underlying to
+    /// @param owner Account to burn pool shares from
+    /// @return assets Amount of underlying withdrawn
+    function redeem(uint256 shares, address receiver, address owner) external override returns (uint256 assets);
+
+    /// @notice Burns caller's pool shares
+    /// @param shares Number of shares to burn
     function burn(uint256 shares) external;
```
0xmikko: Is it safe to keep `burn` public?

contracts/interfaces/IPoolV3.sol
```diff
@@ -1,127 +1,259 @@
...
-    function creditManagerLimit(address) external view returns (uint256);
+    /// @notice Interest rate model contract address
+    function interestRateModel() external view returns (address);
 
-    /// @dev Whether the pool supports quotas
+    /// @notice Annual interest rate in ray that credit account owners pay per unit of borrowed capital
+    function baseInterestRate() external view returns (uint256);
+
+    /// @notice Current cumulative base interest index in ray
+    function baseInterestIndex() external view returns (uint256);
```
0xmikko, line 197: baseCumulativeInterestIndex ?

contracts/interfaces/IPoolV3.sol
```diff
@@ -1,127 +1,259 @@
...
+
+    /// @dev Same as `baseInterestIndex`, kept for backward compatibility
+    function calcLinearCumulative_RAY() external view override returns (uint256);
+
+    /// @notice Cumulative base interest index stored as of last update in ray
+    function baseInterestIndexStored() external view returns (uint256);
```
0xmikko: Stored => LastUpdate for consistency

contracts/interfaces/IPoolV3.sol
```diff
@@ -1,127 +1,259 @@
...
+
+    /// @notice Timestamp of the last base interest rate and index update
+    function lastBaseInterestUpdate() external view returns (uint40);
```
0xmikko, line 206: baseInterestLastUpdate

contracts/interfaces/IPoolV3.sol
```diff
@@ -1,127 +1,259 @@
...
 
-    /// @dev Timestamp of the pool's last update
-    function timestampLU() external view returns (uint64);
+    /// @notice Timestamp of the last quota revenue update
+    function lastQuotaRevenueUpdate() external view returns (uint40);
 
-    function totalBorrowedLimit() external view returns (uint256);
+    /// @notice Updates quota revenue value by given delta
+    /// @param quotaRevenueDelta Quota revenue delta
+    function updateQuotaRevenue(int256 quotaRevenueDelta) external;
```
0xmikko, line 226: why int256 ?

contracts/interfaces/IPoolV3.sol
```diff
@@ -1,127 +1,259 @@
...
 
-    /// @dev Address provider
-    function addressProvider() external view returns (address);
+    /// @notice Sets new quota revenue value
+    /// @param newQuotaRevenue New quota revenue value
+    function setQuotaRevenue(uint256 newQuotaRevenue) external;
```
0xmikko, line 230: why uint256?

contracts/pool/PoolV3.sol
```diff
@@ -1,771 +1,693 @@
...
+        ERC4626(IERC20(underlyingToken_))
         ERC20(
-            string(
-                abi.encodePacked("diesel ", _underlyingToken != address(0) ? IERC20Metadata(_underlyingToken).name() : "")
-            ),
-            string(abi.encodePacked("d", _underlyingToken != address(0) ? IERC20Metadata(_underlyingToken).symbol() : ""))
-        ) // F:[P4-01]
-        nonZeroAddress(_underlyingToken) // F:[P4-02]
-        nonZeroAddress(_interestRateModel) // F:[P4-02]
+            string(abi.encodePacked("diesel ", underlyingToken_ != address(0) ? ERC20(underlyingToken_).name() : "")),
```
0xmikko: how to solve 2 alpha pool issue?

contracts/pool/PoolV3.sol
```diff
@@ -1,771 +1,693 @@
...
-    /// As opposed to {deposit}, minting is allowed even if the vault is in a state where the price of a share is zero.
-    /// In this case, the shares will be minted without requiring any assets to be deposited.
+    /// @inheritdoc IPoolV3
     function mint(uint256 shares, address receiver)
         public
-        override(ERC4626, IERC4626)
-        whenNotPaused // F:[P4-4]
+        override(ERC4626, IPoolV3)
+        whenNotPaused // U:[P4-4]
         nonReentrant
```
0xmikko, line 215: nonReentrant should be covered with tests

contracts/pool/PoolV3.sol
```diff
@@ -1,771 +1,693 @@
...
-        }
-
-        emit Withdraw(msg.sender, receiver, owner, assetsDelivered, shares); // F:[P4-8, 9]
+        uint256 assetsSent = _convertToAssets(shares, Math.Rounding.Down); // U:[P4-9]
+        assets = _amountMinusFee(_amountMinusWithdrawalFee(assetsSent)); // U:[P4-9]
+        _withdraw(receiver, owner, assetsSent, assets, shares); // U:[P4-9]
     }
 
+    /// @inheritdoc IPoolV3
     function burn(uint256 shares)
```
0xmikko: Let's make it creditConfiguratorOnly

contracts/pool/PoolV3.sol
```diff
@@ -1,771 +1,693 @@
...
+    function previewRedeem(uint256 shares) public view override(ERC4626, IPoolV3) returns (uint256) {
+        return _amountMinusFee(_amountMinusWithdrawalFee(_convertToAssets(shares, Math.Rounding.Down)));
     }
 
-    /// @dev See {IERC4626-previewDeposit}.
-    function previewDeposit(uint256 assets) public view override(ERC4626, IERC4626) returns (uint256) {
-        return _convertToShares(_amountMinusFee(assets), Math.Rounding.Down); // TODO: add fee parameter
+    /// @inheritdoc IPoolV3
+    function maxDeposit(address) public view override(ERC4626, IPoolV3) returns (uint256) {
+        return paused() ? 0 : type(uint256).max;
```
0xmikko, line 274: Could it has practical value to provide real maxDeposited limited with current diesel rate and availbale tokens in the pool?

contracts/pool/PoolV3.sol
```diff
@@ -1,771 +1,693 @@
...
     }
 
-    /// @dev See {IERC4626-maxMint}.
-    function maxMint(address) public view override(ERC4626, IERC4626) returns (uint256) {
-        uint128 limit = expectedLiquidityLimit128;
-        return (limit == type(uint128).max) ? type(uint256).max : previewMint(limit - expectedLiquidity());
+    /// @inheritdoc IPoolV3
+    function maxMint(address) public view override(ERC4626, IPoolV3) returns (uint256) {
```
0xmikko, line 278: Could it has practical value to provide real maxDeposited limited with current diesel rate and availbale tokens in the pool?

0xmikko (approved):

contracts/interfaces/IPoolV3.sol
lekhovitsky: General idea: work with full types everywhere, and only cast them when saving to the storage

It's cheaper in terms of gas and reduces the casting-related fuckery across the code that doesn't store anything

contracts/pool/PoolV3.sol
lekhovitsky: From ERC-4626 [spec](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/interfaces/IERC4626.sol#L73-L81): 

>      - MUST return 2 ** 256 - 1 if there is no limit on the maximum amount of assets that may be deposited.

So, yeah, technically the true limit is `type(uint128).max - expectedLiquidity()`, but that's still a huge number so might as well return "infinity".

contracts/interfaces/IPoolV3.sol
lekhovitsky: I would be genuinely confused to see both `baseInterestLastUpdate` (which is the last base interest stored in the contract) and `lastBaseInterestUpdate` (which is the timestamp of the last update)

And the `Stored` naming matches that of [Compound](https://github.com/compound-finance/compound-protocol/blob/master/contracts/CToken.sol#L270-L286)

contracts/interfaces/IPoolV3.sol
lekhovitsky: done

contracts/pool/PoolV3.sol
lekhovitsky: Well, symbol `ddUSDC` looks fine, name `diesel diesel USD coin` doesn't. Not sure, if anything can be done about that except setting the name and the symbol manually via constructor arguments.

contracts/interfaces/IPoolV3.sol
lekhovitsky: I guess in this context the name `index` can imply that it's cumulative, especially since there are no other indexes.

And it makes additional sense since `_updateBaseInterest` function updates `baseInterestRate` and `baseInterestIndex`

contracts/pool/PoolV3.sol
lekhovitsky: `treasuryOnly` would probably make more sense, but I don't really want to introduce new roles/modifiers/exceptions, so is it actually possible for configurator to transfer treasury's shares to itself? So that we can make `burn() configuratorOnly`

lekhovitsky: More so, since we're planning to add pools with diesel tokens as underlying, I'd remove the ability for DAO to increase its price

contracts/interfaces/IPoolV3.sol
lekhovitsky: removed

contracts/interfaces/IPoolV3.sol
lekhovitsky: Changes `Stored` -> `LU` across the code

0xmikko (approved):

PR #53, opened 2023-05-29
fix: updateRate bugfix + quotas logic test

Van0k:


contracts/pool/PoolQuotaKeeperV3.sol
```diff
@@ -315,7 +315,7 @@ contract PoolQuotaKeeperV3 is IPoolQuotaKeeperV3, ACLNonReentrantTrait, Contract
             /// Before writing a new rate, the token's interest index current value is also
             /// saved, to ensure that further calculations with the new rates are correct
             TokenQuotaParams storage tokenQuotaParams = totalQuotaParams[token];
-            quotaRevenue += tokenQuotaParams.updateRate(timeFromLastUpdate, rate);
+            quotaRevenue += tokenQuotaParams.updateRate(timestampLU, rate);
```
0xmikko, line 318: why lastQuotaRateUpdate doesn't use as paramter directly?

contracts/libraries/QuotasLogic.sol
```diff
@@ -75,7 +75,7 @@ library QuotasLogic {
         pure
         returns (uint256)
     {
-        return quoted * (cumulativeIndexNow - cumulativeIndexLU) / RAY;
+        return quoted * (cumulativeIndexNow - cumulativeIndexLU) / RAY; // U: [QL-2]
```
0xmikko: please, wrap quoted with uint256() to eliminate overflow, otherwise if quoted * RAY is greater than uint96 there could be some problems there

contracts/libraries/QuotasLogic.sol
```diff
@@ -147,53 +147,53 @@ library QuotasLogic {
...
-            accountQuota.quota += change;
-            tokenQuotaParams.totalQuoted += change;
+            accountQuota.quota += change; // U: [QL-3]
+            tokenQuotaParams.totalQuoted += change; // U: [QL-3]
 
             // For some tokens, a one-time quota increase fee may be charged. This is a proxy for
             // trading fees for tokens with high volume but short position duration, in which
             // case trading fees are a more effective pricing policy than charging interest over time
-            caQuotaInterestChange += change * tokenQuotaParams.quotaIncreaseFee / PERCENTAGE_FACTOR;
+            caQuotaInterestChange += change * tokenQuotaParams.quotaIncreaseFee / PERCENTAGE_FACTOR; // U: [QL-3]
```
0xmikko: It's also preferable to wrap change with caQIC uint size first

0xmikko (approved):

PR #57, opened 2023-05-30
feat: trading fees solution

0xmikko:


contracts/interfaces/ICreditManagerV3.sol
```diff
@@ -44,7 +45,8 @@ struct CollateralDebtData {
     uint256 debt;
     uint256 cumulativeIndexNow;
     uint256 cumulativeIndexLastUpdate;
-    uint256 cumulativeQuotaInterest;
+    uint128 cumulativeQuotaInterest;
+    uint128 quotaProfits;
```
lekhovitsky, line 49: Since `CollateralDebtData` only lives in memory, it's better to keep these two as `uint256`

contracts/libraries/CreditLogic.sol
```diff
@@ -227,22 +230,42 @@ library CreditLogic {
...
+            uint256 profit,
+            uint128 newCumulativeQuotaInterest,
+            uint128 newQuotaProfits
+        )
     {
         uint256 amountToRepay = amount;
 
-        /// The debt is repaid in the order of: quota interest -> base interest -> debt
-        /// I.e., if the repayment amount only partially covers quota interest, then that will be
+        /// The debt is repaid in the order of: quota profits -> quota interest -> base interest -> debt
```
lekhovitsky, line 249: shouldn't base interest fee be mentioned somewhere in this chain?

contracts/interfaces/ICreditManagerV3.sol
```diff
@@ -23,7 +23,8 @@ uint8 constant BOT_PERMISSIONS_SET_FLAG = 1 << 1;
 struct CreditAccountInfo {
     uint256 debt;
     uint256 cumulativeIndexLastUpdate;
-    uint256 cumulativeQuotaInterest;
+    uint128 cumulativeQuotaInterest;
+    uint128 quotaProfits;
```
lekhovitsky, line 27: Terminology is a bit inconsistent
* `tradingFees` in PQK vs `quotaProfits` here
* we're calling them **profits** in `CreditAccountInfo`, but they are, in fact, **paid by** credit accounts

Maybe name them `quotaFees` or `tradingFees` everywhere?

contracts/credit/CreditManagerV3.sol
```diff
@@ -502,6 +506,7 @@ contract CreditManagerV3 is ICreditManagerV3, SanityCheckTrait, ReentrancyGuardT
                     tokens: collateralDebtData.quotedTokens
                 });
                 creditAccountInfo[creditAccount].cumulativeQuotaInterest = newCumulativeQuotaInterest + 1; // U:[CM-11]
+                creditAccountInfo[creditAccount].quotaProfits = newQuotaProfits + 1;
```
lekhovitsky, line 509: I think it must be `max(x, 1)` instead of `x + 1` on both lines

contracts/libraries/CreditLogic.sol
```diff
@@ -227,22 +230,42 @@ library CreditLogic {
...
+        /// to repay quota interest, and so on. If the repayment amount only partially covers quota interest, then that will be
         /// partially repaid (with part of payment going to fees pro-rata), while base interest
         /// and debt remain inchanged. If the amount covers quota interest fully, then the same logic
         /// applies to the remaining amount and base interest/debt.
 
-        if (cumulativeQuotaInterest != 0) {
+        if (quotaProfits != 0) {
+            if (amountToRepay > quotaProfits) {
+                newQuotaProfits = 0;
+                amountToRepay -= quotaProfits;
```
lekhovitsky, line 259: may be `unchecked`

lekhovitsky (approved):

contracts/credit/CreditManagerV3.sol
0xmikko: +1 should be removed, because the slot is already warmed with `cumulativeQuotaInterest`

contracts/libraries/CreditLogic.sol
0xmikko: Done

PR #61, opened 2023-05-30
feat: combine wm and wg, refactor cc

lekhovitsky:


contracts/support/WithdrawalManagerV3.sol
```diff
@@ -88,15 +105,25 @@ contract WithdrawalManagerV3 is IWithdrawalManagerV3, ACLTrait, ContractsRegiste
     }
 
     /// @dev Sends all `account`'s immediately withdrawable balance of `token` to `to`
-    function _claimImmediateWithdrawal(address account, address token, address to) internal {
+    function _claimImmediateWithdrawal(address account, address token, address to, bool unwrapWETH) internal {
```
0xmikko, line 115: Original idea was to add address 0xeeeee... for ETH to remove annoying flag convert to ETH

contracts/credit/CreditManagerV3.sol
```diff
@@ -1124,22 +1119,26 @@ contract CreditManagerV3 is ICreditManagerV3, SanityCheckTrait, ReentrancyGuardT
...
+    /// @param convertToETH If true, WETH will be transferred to withdrawal manager, from which Credit Facade can
+    ///        claim it as ETH later in the transaction (ignored if token is not WETH)
     function _safeTokenTransfer(address creditAccount, address token, address to, uint256 amount, bool convertToETH)
         internal
     {
         if (convertToETH && token == weth) {
-            ICreditAccountBase(creditAccount).transfer({token: token, to: wethGateway, amount: amount}); // U:[CM-31, 32]
-            IWETHGatewayV3(wethGateway).deposit({to: to, amount: amount}); // U:[CM-31, 32]
+            ICreditAccountBase(creditAccount).transfer({token: token, to: withdrawalManager, amount: amount}); // U:[CM-31, 32]
+            IWithdrawalManagerV3(withdrawalManager).addImmediateWithdrawal({
```
0xmikko: Please, add a helper function for addImmediteWithdrawal for contract size reduction

contracts/credit/CreditConfiguratorV3.sol
```diff
@@ -156,19 +153,19 @@ contract CreditConfiguratorV3 is ICreditConfiguratorV3, ACLNonReentrantTrait {
...
             emit SetPriceOracle(CreditManagerV3(creditManager).priceOracle()); // I:[CC-1A]
 
-            /// Sets the max debt per block multiplier
-            /// This parameter determines the maximal new debt per block as a factor of
-            /// maximal Credit Account debt - essentially a cap on the number of new Credit Accounts per block
-            _setMaxDebtPerBlockMultiplier(uint8(DEFAULT_LIMIT_PER_BLOCK_MULTIPLIER)); // I:[CC-1]
+            // Sets the max debt per block multiplier
+            // This parameter determines the maximal new debt per block as a factor of
+            // maximal Credit Account debt - essentially a cap on the number of new Credit Accounts per block
+            _setMaxDebtPerBlockMultiplier(address(_creditFacade), uint8(DEFAULT_LIMIT_PER_BLOCK_MULTIPLIER)); // I:[CC-1]
```
0xmikko, line 165: What the motivation to add creditFacade as parameter?

lekhovitsky: In the original implementation, `creditFacade()` was called multiple times in every function, which added around 4-5K gas to the `setCreditFacade`

contracts/support/WithdrawalManagerV3.sol
lekhovitsky: Well, yeah, that's neater, will do

contracts/credit/CreditManagerV3.sol
lekhovitsky: Done

contracts/support/WithdrawalManagerV3.sol
```diff
@@ -88,12 +90,17 @@ contract WithdrawalManagerV3 is IWithdrawalManagerV3, ACLTrait, ContractsRegiste
...
-    function claimImmediateWithdrawal(address token, address to, bool unwrapWETH)
+    function claimImmediateWithdrawal(address token, address to)
         external
         override
         nonZeroAddress(to) // U:[WM-4A]
     {
-        _claimImmediateWithdrawal({account: msg.sender, token: token, to: to, unwrapWETH: unwrapWETH});
+        _claimImmediateWithdrawal({
+            account: msg.sender,
+            token: token == ETH_ADDRESS ? weth : token,
```
0xmikko, line 100: Is it correct, it seems that token balance will be stored under ETH_ADDRESS index in map, and it's good to separate it with WETH (user could also have some WETH before). My suggestion is to keep balance under ETH_ADDRESS  as token, and withdraw WETH and send ETH in `_claim...` function

lekhovitsky: `ETH_ADDRESS` only matters during claim, `addImmediateWithdrawal` and mapping are unaware of it

0xmikko (approved):

contracts/support/WithdrawalManagerV3.sol
0xmikko: Got your point, I though it's added in CM under ETH_ADDRESS

PR #65, opened 2023-05-31
fix: botlist fixes and cm optimisation

0xmikko:


contracts/support/BotListV3.sol
```diff
@@ -266,6 +265,14 @@ contract BotListV3 is ACLNonReentrantTrait, IBotListV3 {
         }
     }
 
+    /// @notice Reverts if caller is not creditFacade
+    function _revertIfNotValidCreditFacade() internal view {
```
lekhovitsky: Rename to `_revertIfCallerNotValidCreditFacade` or add `caller` as argument

contracts/libraries/CreditLogic.sol
```diff
@@ -313,8 +317,6 @@ library CreditLogic {
                 profit += amountToRepay - amountToPool; // U:[CL-03B]
                 amountToRepay = 0; // U:[CL-03B]
 
-                newDebt = debt; // U:[CL-03A]
```
lekhovitsky, line 316: Probably this line can also be removed from the quotas block

lekhovitsky (approved):

contracts/libraries/CreditLogic.sol
0xmikko: It's removed, how you found there?

PR #66, opened 2023-05-31
fix: check & optimisation

0xmikko:


lekhovitsky (approved):
please check for compiler warnings related to this PR and remove `console` imports

I'd also suggest renaming `limit` argument of `calcCollateral` into `targetCollateral` or `targetTWV`

nice optimizations, otherwise

PR #70, opened 2023-06-02
fix: remove exceptions from balance logic library

lekhovitsky:


contracts/credit/CreditFacadeV3.sol
```diff
@@ -824,7 +824,8 @@ contract CreditFacadeV3 is ICreditFacadeV3, ACLNonReentrantTrait {
         // If expectedBalances was set by calling revertIfGetLessThan,
         // checks that actual token balances are not less than expected balances
         if (expectedBalances.length != 0) {
-            BalancesLogic.compareBalances(creditAccount, expectedBalances); // U:[FA-23]
+            bool violation = BalancesLogic.compareBalances(creditAccount, expectedBalances);
```
0xmikko: as for me, `bool success` is more common practise

contracts/libraries/BalancesLogic.sol
```diff
@@ -1,10 +1,9 @@
 // SPDX-License-Identifier: MIT
 // Gearbox Protocol. Generalized leverage for DeFi protocols
-// (c) Gearbox Holdings, 2022
+// (c) Gearbox Holdings, 2023
```
0xmikko, line 3: Maybe, it's better to update the whole codebase with one action?

contracts/credit/CreditFacadeV3.sol
lekhovitsky: makes sense, but "if (success) revert" seems confusing :)

contracts/libraries/BalancesLogic.sol
lekhovitsky: let's call it a built-in "format-on-save" hook in my brain

contracts/credit/CreditFacadeV3.sol
0xmikko: what if....? I suggest...
```
 bool success = BalancesLogic.compareBalances(creditAccount, expectedBalances);
 if (!success) revert(...)
 ```

**Review by 0xmikko — APPROVED** [review]

File: contracts/libraries/BalancesLogic.sol [hunk 3]
```diff
@@ -30,30 +28,30 @@ library BalancesLogic {
...
         }
     }
 
     /// @dev Compares current balances to previously saved expected balances.
-    /// Reverts if at least one balance is lower than expected
     /// @param creditAccount Credit Account to check
     /// @param expected Expected balances after all operations
-    function compareBalances(address creditAccount, Balance[] memory expected) internal view {
-        uint256 len = expected.length; // F:[FA-45]
+    /// @return False if at least one balance is lower than expected, true otherwise
```
0xmikko: Probably, named return could help others to understand the logic

0xmikko (approved):

contracts/libraries/BalancesLogic.sol
lekhovitsky: done

PR #75, opened 2023-06-05
fix: remove price feed type interface

lekhovitsky:


contracts/test/mocks/oracles/PriceFeedMock.sol
```diff
@@ -4,7 +4,7 @@
 pragma solidity ^0.8.10;
 
 import {AggregatorV3Interface} from "@chainlink/contracts/src/v0.8/interfaces/AggregatorV3Interface.sol";
-import {IPriceFeedType, PriceFeedType} from "../../../interfaces/IPriceFeedType.sol";
+import {IPriceFeedType, PriceFeedType} from "@gearbox-protocol/core-v2/contracts/interfaces/IPriceFeedType.sol";
```
0xmikko: incorrect npm, it should be `integration-types` I think

0xmikko (approved):

PR #76, opened 2023-06-05
feat: controller rework

Van0k:
Controller-related logic reworked:
1) `controllerOnly` functions now allow access to configurator regardless of whether the controller is set
2) PolicyManager now has `admin` field in `Policy` and checks it in `checkPolicy`. This allows to configure admins per configured parameter, which enables having different functions in the same contract (i.e. CreditConfigurator) under different admins. The execution of a transaction is always performed by the address that scheduled it, i.e. the admin in the corresponding policy

### DIFF
--- a/contracts/interfaces/IControllerTimelockV3.sol
+++ b/contracts/interfaces/IControllerTimelockV3.sol
@@ -21,7 +21,7 @@ interface IControllerTimelockV3Events {
 
     /// @dev Emits when a transaction is queued
     event QueueTransaction(
-        bytes32 indexed txHash, address executor, address target, string signature, bytes data, uint40 eta
+        bytes32 indexed txHash, address indexed executor, address target, string signature, bytes data, uint40 eta
     );
 
     /// @dev Emits when a transaction is executed
--- a/contracts/interfaces/IControllerTimelockV3.sol
+++ b/contracts/interfaces/IControllerTimelockV3.sol
@@ -5,6 +5,7 @@ pragma solidity ^0.8.17;
 
 struct QueuedTransactionData {
     bool queued;
+    address executor;
     address target;
     uint40 eta;
     string signature;
@@ -12,9 +13,6 @@ struct QueuedTransactionData {
 }
 
 interface IControllerTimelockV3Events {
-    /// @dev Emits when the admin of the controller is updated
-    event SetAdmin(address indexed newAdmin);
-
     /// @dev Emits when the veto admin of the controller is updated
     event SetVetoAdmin(address indexed newAdmin);
 
@@ -22,7 +20,9 @@ interface IControllerTimelockV3Events {
     event SetDelay(uint256 newDelay);
 
     /// @dev Emits when a transaction is queued
-    event QueueTransaction(bytes32 indexed txHash, address target, string signature, bytes data, uint40 eta);
+    event QueueTransaction(
+        bytes32 indexed txHash, address executor, address target, string signature, bytes data, uint40 eta
+    );
 
     /// @dev Emits when a transaction is executed
     event ExecuteTransaction(bytes32 indexed txHash);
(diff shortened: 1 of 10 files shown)


Van0k:
changeQuotaTokenMinRate removed from Gauge, as it is redundant after the rework

contracts/traits/ReentrancyGuardTrait.sol
```diff
@@ -29,4 +29,10 @@ abstract contract ReentrancyGuardTrait {
         // https://eips.ethereum.org/EIPS/eip-2200)
         _reentrancyStatus = NOT_ENTERED;
     }
+
+    /// @dev Reverts if the contract is currently entered
+    /// @dev /// @dev Used to cut contract size on modifiers
+    function _ensureNotEntered() internal view {
```
lekhovitsky, line 35: is size reduction significant in this case?

because the nice "feature" of this contract was that it was an exact copy of OZ's implementation except slot size, so it might have been excluded from the audit scope

contracts/interfaces/IControllerTimelockV3.sol
```diff
@@ -5,24 +5,24 @@ pragma solidity ^0.8.17;
...
     /// @dev Emits when the veto admin of the controller is updated
     event SetVetoAdmin(address indexed newAdmin);
 
     /// @dev Emits when the delay is changed
     event SetDelay(uint256 newDelay);
 
     /// @dev Emits when a transaction is queued
-    event QueueTransaction(bytes32 indexed txHash, address target, string signature, bytes data, uint40 eta);
+    event QueueTransaction(
+        bytes32 indexed txHash, address executor, address target, string signature, bytes data, uint40 eta
```
lekhovitsky: maybe let's make `executor` and `target` indexed as well?

contracts/traits/ACLTrait.sol
```diff
@@ -24,10 +24,16 @@ abstract contract ACLTrait is SanityCheckTrait {
 
     /// @dev Ensures that function caller has configurator role
     modifier configuratorOnly() {
+        _ensureConfigurator();
+        _;
+    }
+
+    /// @dev Reverts if the caller is not the configurator
+    /// @dev Used to cut contract size on modifiers
+    function _ensureConfigurator() internal view {
```
lekhovitsky: wdyt about renaming this into `_ensureCallerIsConfigurator` for verbosity? same for `ACLNonReentrantTrait`'s methods

contracts/traits/ACLNonReentrantTrait.sol
```diff
@@ -76,7 +85,6 @@ abstract contract ACLNonReentrantTrait is ACLTrait, Pausable, ReentrancyGuardTra
 
     /// @notice Sets new external controller, can only be called by configurator
     function setController(address newController) external configuratorOnly {
-        externalController = !_isConfigurator(newController);
         controller = newController;
```
lekhovitsky, line 89: it might be worth adding a check `if (newController == controller) return;` to avoid redundant events (and for consistency with other similar functions)

contracts/traits/ReentrancyGuardTrait.sol
```diff
@@ -29,4 +29,10 @@ abstract contract ReentrancyGuardTrait {
         // https://eips.ethereum.org/EIPS/eip-2200)
         _reentrancyStatus = NOT_ENTERED;
     }
+
+    /// @dev Reverts if the contract is currently entered
+    /// @dev /// @dev Used to cut contract size on modifiers
```
lekhovitsky: something wrong with comment

lekhovitsky (approved):

contracts/traits/ReentrancyGuardTrait.sol
0xmikko: Also agree  this size reduction is redundant

contracts/interfaces/IControllerTimelockV3.sol
0xmikko: make sense, we do not try to save gas here

contracts/traits/ReentrancyGuardTrait.sol
Van0k: It differs from contract to contract, but quite significant overall
CreditManagerV3: 23.843 -> 22.948
CreditFacadeV3: 22.227 -> 21.761

contracts/interfaces/IControllerTimelockV3.sol
Van0k: It seems to me that if we add target to indexed, we also need to add the signature (since the target/signature pair represents a unique parameter being configured usually). We can't add both, so I'd keep them in data and add only the executor

PR #83, opened 2023-06-15
test: integration tests setup fixed

0xmikko:


contracts/credit/CreditConfiguratorV3.sol
```diff
@@ -211,7 +211,7 @@ contract CreditConfiguratorV3 is ICreditConfiguratorV3, ACLNonReentrantTrait {
         // сreditManager has an additional check that the token is not added yet
         CreditManagerV3(creditManager).addToken({token: token}); // I:[CC-4]
 
-        if (_isQuotedToken(token)) {
+        if (CreditManagerV3(creditManager).supportsQuotas() && _isQuotedToken(token)) {
```
lekhovitsky, line 214: Even when credit manager supports quotas, pool quota keeper can be a zero address, which would still result in a revert. So, in `_isQuotedToken`, it's worth to add the following check:
```solidity
if (quotaKeeper == address(0)) return false;
```
The `supportsQuotas` check on L214 can then be removed.

Related, in the external function `makeTokenQuoted`, it's actually worth to add the `supportsQuotas` check before even checking whether the token is quoted:
```solidity
if (!CreditManagerV3(creditManager).supportsQuotas())
    revert QuotasNotSupportedException();
if (!_isQuotedToken(token)) {
    revert TokenIsNotQuotedException();
}
_makeTokenQuoted(token);
```

contracts/pool/PoolV3.sol
```diff
@@ -547,6 +550,9 @@ contract PoolV3 is ERC4626, ACLNonReentrantTrait, ContractsRegisterTrait, IPoolV
             lastBaseInterestUpdate = uint40(block.timestamp); // U:[P4-16]
         }
 
+        expectedLiquidity_ = (expectedLiquidity_.toInt256() + expectedLiquidityDelta).toUint256();
+        uint256 availableLiquidity_ = (availableLiquidity().toInt256() + availableLiquidityDelta).toUint256(); // U:[P4-16]
+
```
lekhovitsky: let's fix this in another PR

lekhovitsky (changes requested):

lekhovitsky (approved):

PR #84, opened 2023-06-18
fix: chainsecurity audit findings fixes

lekhovitsky:


contracts/core/AccountFactoryV3.sol
```diff
@@ -52,7 +52,7 @@ contract AccountFactoryV3 is IAccountFactoryV3, ACLTrait, ContractsRegisterTrait
     mapping(address => FactoryParams) internal _factoryParams;
 
     /// @dev Mapping credit manager => queued accounts
-    mapping(address => QueuedAccount[]) internal _queuedAccounts;
+    mapping(address => QueuedAccount[2 ** 32]) internal _queuedAccounts;
```
0xmikko, line 55: Let's add a comment, that we should not use length function for that. It's quite intesresting trick

0xmikko (approved):

contracts/core/AccountFactoryV3.sol
0xmikko: It's not good solution, the correct approach is to change it to mapping(uint256 => QueuedAccount) which will be much more gas efficient

lekhovitsky: gas consumption of those is roughly the same (option with mapping consumes additional gas cost of 30 to compute an extra keccak, and option with array spends some gas to check whether index is in bounds)

PR #86, opened 2023-06-23
fix: fixes to various findings

lekhovitsky:


contracts/credit/CreditFacadeV3.sol
```diff
@@ -674,7 +684,7 @@ contract CreditFacadeV3 is ICreditFacadeV3, ACLNonReentrantTrait {
                     else if (method == ICreditFacadeV3Multicall.updateQuota.selector) {
                         _revertIfNoPermission(flags, UPDATE_QUOTA_PERMISSION); // U:[FA-21]
                         (uint256 tokensToEnable, uint256 tokensToDisable) =
-                            _updateQuota(creditAccount, mcall.callData[4:]); // U:[FA-34]
+                            _updateQuota(creditAccount, mcall.callData[4:], flags & FORBIDDEN_TOKENS_ON_ACCOUNT > 0); // U:[FA-34]
```
0xmikko: !=0 is better than > 0

contracts/credit/CreditFacadeV3.sol
```diff
@@ -686,6 +696,9 @@ contract CreditFacadeV3 is ICreditFacadeV3, ACLNonReentrantTrait {
                     /// the account owner can claim the withdrawal.
                     else if (method == ICreditFacadeV3Multicall.scheduleWithdrawal.selector) {
                         _revertIfNoPermission(flags, WITHDRAW_PERMISSION); // U:[FA-21]
+
+                        flags = flags.enable(REVERT_ON_FORBIDDEN_TOKENS);
```
0xmikko: Can we find a better naming than "REVERT_ON_FORBIDDEN_TOKENS"? It doesnt represent the idea

contracts/credit/CreditFacadeV3.sol
```diff
@@ -974,11 +987,17 @@ contract CreditFacadeV3 is ICreditFacadeV3, ACLNonReentrantTrait {
     /// @notice Requests Credit Manager to update a Credit Account's quota for a certain token
     /// @param creditAccount Credit Account to update the quota for
     /// @param callData Bytes calldata for parsing
-    function _updateQuota(address creditAccount, bytes calldata callData)
+    function _updateQuota(address creditAccount, bytes calldata callData, bool hasForbiddenTokens)
```
0xmikko, line 990: is "checkTokenIsForbidden" better naming? It's more related to action that to the fact that account has f. tokens

contracts/governance/GearStakingV3.sol
```diff
@@ -97,21 +101,46 @@ contract GearStakingV3 is ACLNonReentrantTrait, IGearStakingV3 {
...
+    /// @param amount Amount if staked GEAR to migrate
+    /// @param to Address that will receive staked GEAR in the successor contract
+    /// @param votesBefore Votes to apply before sending GEAR to the successor contract
+    /// @param votesAfter Votes to apply in the new contract after sending GEAR
+    function migrate(uint96 amount, address to, MultiVote[] calldata votesBefore, MultiVote[] calldata votesAfter)
+        external
+        nonReentrant
+        nonZeroAddress(successor) // U: [GS-07]
+    {
+        if (votesBefore.length != 0) {
```
0xmikko: could we move length check into _multivote, it'll make code easier

contracts/credit/CreditFacadeV3.sol
lekhovitsky: ✅

contracts/governance/GearStakingV3.sol
lekhovitsky: ✅

contracts/credit/CreditFacadeV3.sol
lekhovitsky: At the end of `_multicall`, we have the following:
```solidity
if ((flags & REVERT_ON_FORBIDDEN_TOKENS != 0) && (enabledTokensMask & forbiddenTokenMask != 0)) {
    revert ForbiddenTokensException(); // U:[FA-27]
}
```

From this snippet, this flag's idea is to indicate whether function should revert if account has any forbidden tokens enabled, so it would probably be hard to come up with a better name

contracts/credit/CreditFacadeV3.sol
lekhovitsky: Technically, `checkTokenIsForbidden = hasForbiddenTokens && quotaChange > 0`, so the proper naming becomes complicated... Let's just add a clear comment on what this variable means.

contracts/credit/CreditFacadeV3.sol
```diff
@@ -566,7 +569,12 @@ contract CreditFacadeV3 is ICreditFacadeV3, ACLNonReentrantTrait {
             getTokenByMaskFn: _getTokenByMask
         });
 
-        FullCheckParams memory fullCheckParams = _multicall(creditAccount, calls, enabledTokensMaskBefore, permissions);
+        FullCheckParams memory fullCheckParams = _multicall(
+            creditAccount,
+            calls,
+            enabledTokensMaskBefore,
+            permissions | (forbiddenBalances.length != 0 ? FORBIDDEN_TOKENS_ON_ACCOUNT : 0)
```
0xmikko: it's better to use enable from BitMask lib

contracts/interfaces/ICreditFacadeV3Multicall.sol
```diff
@@ -24,9 +24,10 @@ uint256 constant ALL_CREDIT_FACADE_CALLS_PERMISSION = ADD_COLLATERAL_PERMISSION
 uint256 constant ALL_PERMISSIONS = ALL_CREDIT_FACADE_CALLS_PERMISSION | EXTERNAL_CALLS_PERMISSION;
 
 // All flags start from 193rd bit, because bot permissions is uint192
-uint256 constant INCREASE_DEBT_WAS_CALLED = 1 << 193;
+uint256 constant REVERT_ON_FORBIDDEN_TOKENS = 1 << 193;
```
0xmikko: better naming? not it looks like that the fuction should revert on ny token, however, the idea is to be reverted if account keeps them at the end of multicall

0xmikko (approved):

contracts/interfaces/ICreditFacadeV3Multicall.sol
lekhovitsky: replied here https://github.com/Gearbox-protocol/core-v3/pull/86#discussion_r1249341031

contracts/credit/CreditFacadeV3.sol
lekhovitsky: ✅

contracts/interfaces/ICreditFacadeV3Multicall.sol
lekhovitsky: renamed to `REVERT_ON_FORBIDDEN_TOKENS_AFTER_CALLS`

PR #91, opened 2023-07-08
feat: bot list special permissions

Van0k:


contracts/credit/CreditFacadeV3.sol
```diff
@@ -534,17 +534,23 @@ contract CreditFacadeV3 is ICreditFacadeV3, ACLNonReentrantTrait {
...
+        // Checks that the bot is approved by the borrower (or has special permissions from DAO) and is not forbidden
+        if (
+            (!hasSpecialPermissions && (_flagsOf(creditAccount) & BOT_PERMISSIONS_SET_FLAG == 0)) || botPermissions == 0
+                || forbidden
+        ) {
             revert NotApprovedBotException(); // U:[FA-19]
         }
 
-        _multicallFullCollateralCheck(creditAccount, calls, botPermissions.enable(PAY_BOT_CAN_BE_CALLED)); // U:[FA-19, 20]
+        botPermissions |= hasSpecialPermissions ? 0 : PAY_BOT_CAN_BE_CALLED;
```
lekhovitsky: can be rewritten via `BitMask`'s `enable` for consistency

contracts/credit/CreditFacadeV3.sol
```diff
@@ -534,17 +534,23 @@ contract CreditFacadeV3 is ICreditFacadeV3, ACLNonReentrantTrait {
...
             bot: msg.sender
         });
-        // Checks that the bot is approved by the borrower and is not forbidden
-        if (_flagsOf(creditAccount) & BOT_PERMISSIONS_SET_FLAG == 0 || botPermissions == 0 || forbidden) {
+
+        // Checks that the bot is approved by the borrower (or has special permissions from DAO) and is not forbidden
+        if (
+            (!hasSpecialPermissions && (_flagsOf(creditAccount) & BOT_PERMISSIONS_SET_FLAG == 0)) || botPermissions == 0
+                || forbidden
+        ) {
```
lekhovitsky: can change the order of operands so that `_flagsOf` is called later, would result in cheaper reverts

contracts/core/BotListV3.sol
```diff
@@ -102,7 +107,12 @@ contract BotListV3 is ACLNonReentrantTrait, IBotListV3 {
...
 
-        if (forbiddenBot[bot] && permissions != 0) {
+        if (
+            (
+                botSpecialStatus[creditManager][bot].forbidden
+                    || botSpecialStatus[creditManager][bot].specialPermissions != 0
+            ) && permissions != 0
+        ) {
             revert InvalidBotException(); // F: [BL-03]
         }
```
lekhovitsky, line 117: In case `specialPermissions != 0`, it probably should revert here even if `permissions == 0`

contracts/core/BotListV3.sol
```diff
@@ -228,19 +238,51 @@ contract BotListV3 is ACLNonReentrantTrait, IBotListV3 {
...
-        returns (uint192 permissions, bool forbidden)
+        returns (uint192 permissions, bool forbidden, bool hasSpecialPermissions)
     {
-        return (botPermissions[creditManager][creditAccount][bot], forbiddenBot[bot]);
+        uint192 specialPermissions;
+        (forbidden, specialPermissions) =
+            (botSpecialStatus[creditManager][bot].forbidden, botSpecialStatus[creditManager][bot].specialPermissions); // F: [BL-7]
+
+        hasSpecialPermissions = specialPermissions != 0;
+        permissions = hasSpecialPermissions ? specialPermissions : botPermissions[creditManager][creditAccount][bot];
```
lekhovitsky, line 248: does it make sense to avoid sloading `botSpecialStatus[creditManager][bot].specialPermissions` and  `botPermissions[creditManager][creditAccount][bot]` in case `forbidden == true` and simply return 0?

contracts/core/BotListV3.sol
```diff
@@ -228,19 +238,51 @@ contract BotListV3 is ACLNonReentrantTrait, IBotListV3 {
...
+        _setBotForbiddenStatus(creditManager, bot, status);
+    }
+
+    /// @notice Sets the bot's forbidden status in all Credit Managers
+    function setBotForbiddenStatusEverywhere(address bot, bool status) external configuratorOnly {
+        uint256 len = approvedCreditManagers.length();
+
+        for (uint256 i = 0; i < len; ++i) {
+            _setBotForbiddenStatus(approvedCreditManagers.at(i), bot, status);
+        }
```
lekhovitsky: can be unchecked

contracts/core/BotListV3.sol
```diff
@@ -228,19 +238,51 @@ contract BotListV3 is ACLNonReentrantTrait, IBotListV3 {
...
+    }
+
+    /// @dev IMPLEMENTATION: setBotForbiddenStatus
+    function _setBotForbiddenStatus(address creditManager, address bot, bool status) internal {
+        botSpecialStatus[creditManager][bot].forbidden = status;
+        emit SetBotForbiddenStatus(creditManager, bot, status);
```
lekhovitsky, line 273: can return early if `botSpecialStatus[creditManager][bot].forbidden == status`

contracts/core/BotListV3.sol
```diff
@@ -228,19 +238,51 @@ contract BotListV3 is ACLNonReentrantTrait, IBotListV3 {
...
+    }
+
+    /// @notice Gives special permissions to a bot that extend to all Credit Accounts
+    /// @dev Bots with special permissions are DAO-approved bots
+    ///      which are enabled with a defined set of permissions for all users.
+    ///      Can be used to extend system functionality with additional features
+    ///      without changing the core - such as adding partial liquidations.
+    function setBotSpecialPermissions(address creditManager, address bot, uint192 permissions)
```
lekhovitsky, line 281: shouldn't we check that `creditManager` is approved?

contracts/core/BotListV3.sol
Van0k: May technically be some edge cases where the user has normal permissions for a bot that is later made special (e.g., there is some community bot the DAO decides to make special at some point). So best to keep the logic the same as forbidden bot logic

contracts/core/BotListV3.sol
Van0k: `specialPermissions` is in the same slot, so hopefully solc will pack both into a single SLOAD. But overall I don't really see cases where a user calls `getBotStatus` for a forbidden bot except when it is a footgun. So don't see any reason to complicate the function for that

contracts/core/BotListV3.sol
Van0k: Decided to keep firing the event always to keep the number of events consistent for `setBotForbiddenStatusEverywhere`. In a critical moment, you don't want to to figure out whether the event wasn't fired because the bot is already forbidden or because something went wrong.

contracts/core/BotListV3.sol
Van0k: May be some cases where we want to prohibit bot list interactions but set special bots regardless.

lekhovitsky (approved):

0xmikko (approved):

PR #95, opened 2023-07-12
feat: gauges and staking migration overhaul

Van0k:


Van0k:
Depositing on behalf of another user can now only be done through a special `depositOnMigration` function, which is restricted to a special `migrator` role. Migrator is set by configurator. Normal deposits no longer accept the 'to' parameter.

contracts/governance/GearStakingV3.sol
```diff
@@ -49,31 +49,44 @@ contract GearStakingV3 is ACLNonReentrantTrait, IGearStakingV3 {
...
     ///              * isIncrease - whether to add or remove votes
     ///              * extraData - data specific to the voting contract that is decoded on recipient side
-    function deposit(uint96 amount, address to, MultiVote[] calldata votes) external nonReentrant {
-        IERC20(gear).safeTransferFrom(msg.sender, address(this), amount); // U: [GS-02]
+    function deposit(uint96 amount, MultiVote[] calldata votes) external nonReentrant {
+        _deposit(amount, msg.sender, msg.sender, votes); // U: [GS-02]
+    }
+
+    /// @dev Internal implementation for deposits. Used in `deposit` and `depositOnMigration`
+    function _deposit(uint96 amount, address from, address to, MultiVote[] calldata votes) internal {
```
lekhovitsky: apparently the `from` parameter can be removed as it's always set to `msg.sender`

lekhovitsky (approved):

contracts/governance/GaugeV3.sol
```diff
@@ -75,6 +78,8 @@ contract GaugeV3 is IGaugeV3, IVotingContractV3, ACLNonReentrantTrait {
 
     /// @dev IMPLEMENTATION: updateEpoch()
     function _checkAndUpdateEpoch() internal {
+        if (epochFrozen) return;
```
0xmikko: I'm not sure, that the best solution. It looks more native to update epoch, but keep the same rates:

```solidity
function _checkAndUpdateEpoch() internal {


        uint16 epochNow = IGearStakingV3(voter).getCurrentEpoch(); // U:[GA-14]

        if (epochNow > epochLastUpdate) {
            epochLastUpdate = epochNow; // U:[GA-14]

            /// The PQK retrieves all rates from the Gauge on its own and saves them
            /// Since this function is only callable by the Gauge, active rates can only
            /// be updated once per epoch at most
            if (!epochFrozen) {  _poolQuotaKeeper().updateRates(); } // U:[GA-14

            emit UpdateEpoch(epochNow); // U:[GA-14]
        }
    }
```

contracts/governance/GaugeV3.sol
```diff
@@ -209,6 +214,18 @@ contract GaugeV3 is IGaugeV3, IVotingContractV3, ACLNonReentrantTrait {
...
 
+    /// @dev Sets the frozen epoch status
+    /// @param status The new status
+    /// @dev The epoch can be frozen to prevent rate updates during periods of instability,
+    ///      e.g., when the Gauges / GearStaking contracts are being migrated
+    function setFrozenEpoch(bool status) external configuratorOnly {
+        if (status != epochFrozen) {
+            epochFrozen = status;
+
+            emit SetFrozenEpoch(status);
```
0xmikko, line 226: The event could also include epoch number from which we freeze or unfreeze rates update. More strict logic could be implemented as only freezing activity to remove any possibility to use this function in different way

contracts/governance/GearStakingV3.sol
```diff
@@ -132,9 +145,21 @@ contract GearStakingV3 is ACLNonReentrantTrait, IGearStakingV3 {
...
+
+        emit MigrateGear(msg.sender, successor, amount); // U: [GS-07]
+    }
 
-        emit MigrateGear(msg.sender, to, successor, amount); // U: [GS-07]
+    /// @notice Performs a deposit on user's behalf from the migrator (usually the previous GearStaking contract)
+    /// @param amount Amount of GEAR to deposit
+    /// @param to Address to deposit to
+    /// @param votes Array of votes to apply after migrating
+    function depositOnMigration(uint96 amount, address to, MultiVote[] calldata votes)
```
0xmikko: I believe it's better to name parameter `onBehalfOf` than `to`

contracts/governance/GaugeV3.sol
lekhovitsky: this way one can delay rates update for up to one epoch: front-run unfreeze transaction with `updateEpoch`, which would update latest epoch and force to wait until the next one

contracts/governance/GearStakingV3.sol
lekhovitsky: ✅

contracts/governance/GaugeV3.sol
lekhovitsky: ✅  (it's expected that `updateEpoch` will still be called regularly, so front-running won't be an issue)

contracts/governance/GaugeV3.sol
lekhovitsky: came to conclusion that lack of unfreezing can cause certain problems if either new or old gauge is flawed

0xmikko (approved):

PR #98, opened 2023-07-28
feat: add `PriceOracleV3`

lekhovitsky:
Compared to the previous version, the new Price Oracle contract:
* fixes incorrect price feed staleness check
* introduces reserve price feeds that can be enabled or disabled under special circumstances


contracts/traits/PriceCheckTrait.sol
```diff
@@ -0,0 +1,20 @@
...
+// Gearbox Protocol. Generalized leverage for DeFi protocols
+// (c) Gearbox Foundation, 2023.
+pragma solidity ^0.8.17;
+
+import {IncorrectPriceException, StalePriceException} from "../interfaces/IExceptions.sol";
+
+/// @title Price check trait
+abstract contract PriceCheckTrait {
+    /// @dev Period since the last update after which the price is considered stale
+    uint256 internal constant STALENESS_PERIOD = 8 hours;
```
lekhovitsky: @0xmikko @Van0k what number can we put here?

Major tokens that we use have a heartbeat of 1 hour, so if the last update was, say, 2 hours ago the price can be considered stale.

For CRV, however, the heartbeat is much longer: https://data.chain.link/ethereum/mainnet/crypto-usd/crv-usd, so with global staleness period we have to tolerate the possibility of major assets being stale for longer time, or otherwise we'll end up with mostly non-functional system

contracts/test/integration/credit/CreditConfigurator.int.t.sol
```diff
@@ -419,23 +419,23 @@ contract CreditConfiguratorIntegrationTest is IntegrationTestHelper, ICreditConf
     function test_I_CC_04_addCollateralToken_adds_new_token_to_creditManager_and_set_lt() public creditTest {
         uint256 tokensCountBefore = creditManager.collateralTokensCount();
 
-        address cLINKToken = tokenTestSuite.addressOf(Tokens.LUNA);
+        address newToken = tokenTestSuite.addressOf(Tokens.wstETH);
```
lekhovitsky, line 422: LUNA removed from SDK, so replaced with wstETH

contracts/traits/PriceCheckTrait.sol
0xmikko: Have no clear answer yet, I think the best approach is to ask auditors opinion

contracts/core/PriceOracleV3.sol
```diff
@@ -0,0 +1,240 @@
...
+        if (!force) {
+            PriceFeedParams memory mainFeedParams = tokenParams.mainFeedParams;
+            (,,, uint256 updatedAt,) = AggregatorV3Interface(mainFeedParams.priceFeed).latestRoundData();
+            if (active != _isStale(updatedAt, mainFeedParams.stalenessPeriod)) {
+                revert ReservePriceFeedStatusNotAllowedException();
+            }
+        }
+
+        tokenParams.useReserve = active;
+        emit SetReservePriceFeedStatus(token, active);
```
lekhovitsky: An unfortunate side-effect in such implementation is that `setReservePriceFeedStatus({token: token, active: false})` can be called after activating the reserve price feed via `forceReservePriceFeedStatus({token: token, active: true})` (e.g., when the price feed starts returning completely wrong values but is not stale).

So, if these are controlled by different roles, there might be a conflict

lekhovitsky: resolved by removing a `force = false` behavior

contracts/core/PriceOracleV3.sol
```diff
@@ -0,0 +1,249 @@
...
+    AddressIsNotContractException,
+    IncorrectTokenContractException,
+    PriceFeedDoesNotExistException
+} from "../interfaces/IExceptions.sol";
+import {IPriceOracleV3, PriceFeedParams} from "../interfaces/IPriceOracleV3.sol";
+
+import {ACLNonReentrantTrait} from "../traits/ACLNonReentrantTrait.sol";
+import {PriceFeedValidationTrait} from "../traits/PriceFeedValidationTrait.sol";
+
+struct PriceFeedConfig {
```
0xmikko: Lets move it into interface

contracts/core/PriceOracleV3.sol
```diff
@@ -0,0 +1,249 @@
...
+            key := keccak256(ptr, 0x1b)
+        }
+    }
+
+    // ------------- //
+    // CONFIGURATION //
+    // ------------- //
+
+    /// @notice Sets price feed for a given token with default staleness period
+    function addPriceFeed(address token, address priceFeed) external override configuratorOnly {
```
0xmikko: Lets remove it, we can usae setPriceFeed which is more relevant

contracts/core/PriceOracleV3.sol
```diff
@@ -0,0 +1,249 @@
...
+    /// @notice Sets `token`'s reserve price feed status to `active`
+    /// @dev Reserve price feed for the token must already be set
+    function setReservePriceFeedStatus(address token, bool active) external override controllerOnly {
+        _setReservePriceFeedStatus(token, active);
+    }
+
+    /// @dev `setPriceFeed` implementation
+    function _setPriceFeed(address token, address priceFeed, uint32 stalenessPeriod) internal {
+        uint8 decimals = _priceFeedsParams[token].decimals;
+
```
0xmikko, line 156: lets add one more line here for better readability:

decimals = decimals == 0 ? _validateToken(token) : decimals;

contracts/core/PriceOracleV3.sol
```diff
@@ -0,0 +1,249 @@
...
+        override
+        configuratorOnly
+    {
+        _setReservePriceFeed(token, priceFeed, stalenessPeriod);
+    }
+
+    /// @notice Sets `token`'s reserve price feed status to `active`
+    /// @dev Reserve price feed for the token must already be set
+    function setReservePriceFeedStatus(address token, bool active) external override controllerOnly {
+        _setReservePriceFeedStatus(token, active);
```
0xmikko: You can add function code here, there is no way to keep in separate as internal

contracts/core/PriceOracleV3.sol
lekhovitsky: done

contracts/core/PriceOracleV3.sol
lekhovitsky: done

contracts/core/PriceOracleV3.sol
lekhovitsky: done

contracts/core/PriceOracleV3.sol
lekhovitsky: it's not really used except constructor and deploy (which loads contract code from this file anyways), so probably it's best to keep it here

0xmikko (approved):

PR #106, opened 2023-08-21
feat: zero debt accounts

Van0k:
- Credit account opening now has considerations for zero debt case
- Transfers to pool are no longer triggered on closing account with zero debt
- manageDebt is skipped for `amount == 0`
- updateQuota with positive `quotaChange` fails with zero debt


contracts/credit/CreditManagerV3.sol
```diff
@@ -48,6 +48,8 @@ import {
 // EXCEPTIONS
 import "../interfaces/IExceptions.sol";
 
+import "forge-std/console.sol";
```
lekhovitsky: artifact

0xmikko (approved):

PR #117, opened 2023-08-28
feat: controller improvements

### LINKED ISSUE #113: Split `GaugeV3.setQuotaRateParams` into two functions + adjust controller accordingly

### LINKED ISSUE #114: Make delays in `ControllerTimelockV3` policy-dependent

Van0k:
1) Transaction delays are now policy-dependent, instead of 1 global delay per contract
2) Two functions `changeQuotaMinRate` and `changeQuotaMaxRate` were added in addition to `changeQuotaTokenRateParams`, and controller functions were adjusted accordingly

### DIFF
--- a/contracts/governance/PolicyManagerV3.sol
+++ b/contracts/governance/PolicyManagerV3.sol
@@ -120,12 +120,12 @@ abstract contract PolicyManagerV3 is ACLNonReentrantTrait {
         return _policies[policyHash].delay;
     }
 
-    /// @dev Returns policy transaction delay, with policy retrieved based on contract and parameter name
+    /// @dev Returns policy transaction delay, with policy retrieved based on policy UID
     function _getPolicyDelay(bytes32 policyHash) internal view returns (uint256) {
         return _policies[policyHash].delay;
     }
 
-    /// @dev Performs parameter checks, with policy retrieved based on policy UID
+    /// @dev Performs parameter checks, with policy retrieved based on contract and parameter name
     function _checkPolicy(address contractAddress, string memory paramName, uint256 oldValue, uint256 newValue)
         internal
         returns (bool)
--- a/contracts/governance/PolicyManagerV3.sol
+++ b/contracts/governance/PolicyManagerV3.sol
@@ -11,6 +11,7 @@ import {PERCENTAGE_FACTOR} from "@gearbox-protocol/core-v2/contracts/libraries/C
 ///         identifier for a set of contracts
 /// @param enabled Determines whether the policy is enabled. A disabled policy will auto-fail the policy check.
 /// @param admin The admin that can change the parameter under the given policy
+/// @param delay The delay before the transaction can be triggered under a given policy
 /// @param flags Bitmask of flags that determine which policy checks to apply on parameter change:
 ///        * 0 - check exact value
 ///        * 1 - check min value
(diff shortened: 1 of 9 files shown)


contracts/governance/PolicyManagerV3.sol
```diff
@@ -112,7 +114,18 @@ abstract contract PolicyManagerV3 is ACLNonReentrantTrait {
...
+        bytes32 policyHash = keccak256(abi.encode(_group[contractAddress], paramName));
+        return _policies[policyHash].delay;
+    }
+
+    /// @dev Returns policy transaction delay, with policy retrieved based on contract and parameter name
+    function _getPolicyDelay(bytes32 policyHash) internal view returns (uint256) {
+        return _policies[policyHash].delay;
+    }
+
+    /// @dev Performs parameter checks, with policy retrieved based on policy UID
```
lekhovitsky: comments got mixed up, should be "based on policy UID" for the second `_getPolicyDelay` and "based on contract and parameter name" for the first `_checkPolicy`

contracts/interfaces/IGaugeV3.sol
```diff
@@ -74,4 +74,8 @@ interface IGaugeV3 is IGaugeV3Events, IVotingContractV3, IVersion {
     function addQuotaToken(address token, uint16 minRate, uint16 maxRate) external;
 
     function changeQuotaTokenRateParams(address token, uint16 minRate, uint16 maxRate) external;
```
lekhovitsky: Can probably get rid of this one at all then

contracts/governance/PolicyManagerV3.sol
```diff
@@ -112,7 +114,18 @@ abstract contract PolicyManagerV3 is ACLNonReentrantTrait {
...
+    /// @dev Returns policy transaction delay, with policy retrieved based on contract and parameter name
+    function _getPolicyDelay(address contractAddress, string memory paramName) internal view returns (uint256) {
+        bytes32 policyHash = keccak256(abi.encode(_group[contractAddress], paramName));
+        return _policies[policyHash].delay;
+    }
+
+    /// @dev Returns policy transaction delay, with policy retrieved based on contract and parameter name
+    function _getPolicyDelay(bytes32 policyHash) internal view returns (uint256) {
+        return _policies[policyHash].delay;
+    }
```
lekhovitsky: Can combine this one with `_checkPolicy` so that `_checkPolicy` returns result and delay, using a smaller number of SLOAD's; that would require a bit more code changes though

contracts/governance/PolicyManagerV3.sol
```diff
@@ -36,6 +37,7 @@ import {PERCENTAGE_FACTOR} from "@gearbox-protocol/core-v2/contracts/libraries/C
 struct Policy {
     bool enabled;
     address admin;
+    uint40 delay;
```
lekhovitsky, line 40: could it make sense to enforce "delay > 0" in `setPolicy`?

contracts/governance/PolicyManagerV3.sol
Van0k: There are cases where the delay is needed for additional checks (such as `rampLiquidationThreshold`), so best to just retrieve this separately. The slot is warm anyway.

contracts/governance/PolicyManagerV3.sol
Van0k: Might be cases where we want to set it to zero, e.g., `forbidAdapter` or `setTokenLimit`, to replicate pausable admin logic for these functions.

lekhovitsky (approved):

PR #125, opened 2023-09-11
fix: external applyPriceOnDemand

0xmikko:


contracts/credit/CreditFacadeV3.sol
```diff
@@ -238,6 +238,9 @@ contract CreditFacadeV3 is ICreditFacadeV3, ACLNonReentrantTrait {
         // Emits an event for Credit Account opening
         emit OpenCreditAccount(creditAccount, onBehalfOf, msg.sender, debt, referralCode); // U:[FA-10]
 
+        // Price feed updates must be applied before the multicall because they affect CA's collateral evaluation
```
lekhovitsky: this comment is redundant everywhere except for liquidation

contracts/credit/CreditFacadeV3.sol
```diff
@@ -918,14 +913,10 @@ contract CreditFacadeV3 is ICreditFacadeV3, ACLNonReentrantTrait {
     /// @dev Should generally be called only when interacting with tokens
     ///         that use on-demand price feeds
     /// @param callData Bytes calldata for parsing
-    function _onDemandPriceUpdate(bytes calldata callData, FullCheckParams memory fullCheckParams) internal {
+    function _onDemandPriceUpdate(bytes calldata callData, address priceOracle) internal {
```
lekhovitsky: The function is now only used once so it can be inlined in `_applyOnDemandPriceUpdates `

contracts/credit/CreditFacadeV3.sol
```diff
@@ -871,22 +866,22 @@ contract CreditFacadeV3 is ICreditFacadeV3, ACLNonReentrantTrait {
 
     /// @dev Applies on-demand price feed updates from the multicall if the are any, returns number of calls remaining
     ///      `onDemandPriceUpdate` calls are expected to be placed before all other calls in the multicall
-    function _applyOnDemandPriceUpdates(MultiCall[] calldata calls) internal returns (uint256 remainingCalls) {
+    function _applyOnDemandPriceUpdates(MultiCall[] calldata calls) internal returns (uint256 skip) {
```
lekhovitsky: `skip` is not a very informative name, consider changing it to `skipCalls`. Besides, the function natspec should be updated to reflect the change, probably to `"returns the number of calls that can be skipped"`.

lekhovitsky (approved):

PR #142, opened 2023-10-11
feat: new GovernorV3 contract

0xmikko:


contracts/governance/GovernorV3.sol
```diff
@@ -0,0 +1,215 @@
...
+}
+
+uint256 constant BATCH_SIZE_BITS = 16;
+
+contract GovernorV3 is IGovernorV3 {
+    using EnumerableSet for EnumerableSet.AddressSet;
+
+    address public immutable override timeLock;
+
+    EnumerableSet.AddressSet internal queueAdmins;
```
lekhovitsky, line 19: lacks a public getter

contracts/governance/GovernorV3.sol
```diff
@@ -0,0 +1,215 @@
...
+        _;
+    }
+
+    modifier vetoAdminOnly() {
+        if (msg.sender != vetoAdmin) revert CallerNotVetoAdminException();
+
+        _;
+    }
+
+    constructor(address _timeLock, address _queueAdmin, address _vetoAdmin) {
```
lekhovitsky, line 46: can we consider renaming `queueAdmin` -> `proposer` and `veroAdmin` -> `canceller`?

contracts/governance/GovernorV3.sol
```diff
@@ -0,0 +1,215 @@
...
+    {
+        return ITimeLock(timeLock).queueTransaction(target, value, signature, data, eta);
+    }
+
+    function sealBatch(TimeLockTx[] calldata txs) external override queueAdminOnly {
+        uint256 len = txs.length;
+        if (len >= 2 ** BATCH_SIZE_BITS) revert IncorrectBatchLengthException();
+
+        ++batchNum;
+        uint240 _batchNum = batchNum;
```
lekhovitsky: Can avoid redundant `SLOAD`:

```solidity
uint240 _batchNum = batchNum;
unchecked {
    ++_batchNum;
}
batchNum = _batchNum;
```

contracts/governance/GovernorV3.sol
```diff
@@ -0,0 +1,215 @@
...
+
+        uint256 batchShifted = uint256(_batchNum) << BATCH_SIZE_BITS;
+
+        unchecked {
+            for (uint256 i = 0; i < len; ++i) {
+                TimeLockTx calldata ttx = txs[i];
+                bytes32 txHash = getTxHash(ttx);
```
lekhovitsky: The following check can prevent sealing a batch with non-queued transactions:

```solidity
if (!ITimeLock(timeLock).queuedTransactions(txHash)) {
    revert TxNotQueuedException();
}
```

contracts/governance/GovernorV3.sol
lekhovitsky: Also, `batchNum` can be initialized to 1 in constructor, and only incremented at the end of `sealBatch`

contracts/governance/GovernorV3.sol
```diff
@@ -0,0 +1,215 @@
...
+
+                if (batchedTransactions[txHash] != 0) {
+                    revert TxHashCollisionException(ttx.target, ttx.value, ttx.signature, ttx.data, ttx.eta);
+                }
+
+                batchedTransactions[txHash] = batchShifted + i;
```
lekhovitsky: replacing with `batchShifted | i` would be a bit more clear

We can even put this into a struct:
```solidity
struct BatchedTx {
    uint240 batchNum;
    uint16 txIndex;
}
```

It may also contain number of txs in the batch (can be zero for all txs except for the first), removing the need for `batchedTransactionsCount`  mapping

contracts/governance/GovernorV3.sol
```diff
@@ -0,0 +1,215 @@
...
+    function executeTransaction(address target, uint256 value, string memory signature, bytes memory data, uint256 eta)
+        external
+        payable
+        override
+        returns (bytes memory)
+    {
+        return _operation(target, value, signature, data, eta, TxAction.Execute);
+    }
+
+    function executeBatch(TimeLockTx[] calldata txs) external payable override {
```
lekhovitsky, line 104: shouldn't we also return an array of results?

contracts/governance/GovernorV3.sol
```diff
@@ -0,0 +1,215 @@
...
+            if (queueAdmins.length() == 1) revert CantRemoveLastQueueAdminException();
+
+            queueAdmins.remove(_admin);
+            emit RemoveQueueAdmin(_admin);
+        }
+    }
+
+    function updateVetoAdmin(address _admin) external override timeLockOnly {
+        _updateVetoAdmin(_admin);
+    }
```
lekhovitsky, line 214: This is a little troublesome in case `vetoAdmin` is malicious.

If we require the call to come from timelock, then it should have been initiated in this contract. But if it is initiated in this contract, it can be cancelled by the malicious `vetoAdmin`.

contracts/governance/GovernorV3.sol
```diff
@@ -0,0 +1,215 @@
...
+        }
+
+        batchedTransactionsCount[_batchNum] = len;
+
+        emit SealBatch(batchNum, len);
+    }
+
+    // EXECUTE
+
+    function executeTransaction(address target, uint256 value, string memory signature, bytes memory data, uint256 eta)
```
lekhovitsky, line 95: `signature` and `data` can be placed in `calldata` (also in `cancelTransaction` and `_operation`)

lekhovitsky (commented):
Looks robust at first glance, although there are some possible optimizations:

* bit magic with batch numbers and tx indices is a bit unnecessary, as it can be replaced with a simple struct; it may also contain number of txs in the batch, eliminating a second mapping
* small concerns regarding `updateVetoAdmin`

contracts/governance/GovernorV3.sol
0xmikko: It will remove human readability, we discussed a lot

contracts/governance/GovernorV3.sol
0xmikko: It's by design, the same logic for configurator role - if it's malicious - everything is broken

contracts/governance/GovernorV3.sol
0xmikko: No ideas what should be returned tbh

contracts/governance/GovernorV3.sol
0xmikko: It seems batchShifter | i  is easier

contracts/governance/GovernorV3.sol
lekhovitsky: nothing to do with readability, `memory` -> `calldata` in function signature

PR #147, opened 2023-10-21
fix: add clarifying comments

lekhovitsky:


contracts/credit/CreditConfiguratorV3.sol
```diff
@@ -884,6 +884,7 @@ contract CreditConfiguratorV3 is ICreditConfiguratorV3, ACLNonReentrantTrait {
             revert AddressIsNotContractException(_contract); // I:[CC-12A,29]
         }
 
+        // any interface with `creditManager()` would work instead of `CreditFacadeV3` here
```
0xmikko, line 888: We assume that contract compatability means that the contract has `creditManager()` function and it points to current CM. CreditFacadeV3() is used here as interface which implements this method only (IAdapter, ICreditConfigurator has this method as well)

0xmikko (approved):

PR #151, opened 2023-10-25
fix: improved liquidation flow

0xmikko:
Account closure and liquidation are separated into two functions in the credit manager.

* Closing an account exists primarily to get all the tokens back on the account at once and return it to the factory, and requires all debt to be repaid.
* The main change in liquidation is in remaining funds logic. Previously, liquidator was required to pay all remaining funds in underlying, which could be quite problematic in case there was large amount of illiquid tokens on the account that counted towards TV but not TWV (withdrawals, amounts not covered by quotas, tokens with zero LT, etc). Now, liquidator has an option to use such tokens to cover remaining funds simply by excluding them from the mask of tokens they would like to receive. They can't however increase balance of such tokens during multicall to prevent them from offloading illiquid tokens onto users in return for liquid ones.
* The other significant change in liquidations is that underlying shortfall (if any) should now be covered via `addCollateral` in multicall instead of a hidden transfer in credit manager.
* Worth noting, liquidation now leaves the account open so that users could handle remaining funds at their will.
* Since after both closure and liquidation the debt is guaranteed to be 0 in `creditAccountInfo` mapping, there is no point in zeroing it out in `openCreditAccount`. Moreover, reading base interest index from the pool is also redundant, as this value is never used (it, however, did cause some division-by-zero errors, which explains changes in the `CreditLogic` library).


contracts/credit/CreditManagerV3.sol
```diff
@@ -802,6 +827,42 @@ contract CreditManagerV3 is ICreditManagerV3, SanityCheckTrait, ReentrancyGuardT
...
+        remainingFunds = underlyingBalance;
+
+        uint256 remainingTokensMask = enabledTokensMask.disable(UNDERLYING_TOKEN_MASK).disable(tokensToTransferMask);
+        if (remainingTokensMask == 0) return (remainingFunds, underlyingBalance);
+
+        address _priceOracle = priceOracle;
+        uint256 totalValueUSD;
+        while (remainingTokensMask != 0) {
+            uint256 tokenMask = remainingTokensMask & uint256(-int256(remainingTokensMask));
+            remainingTokensMask ^= tokenMask;
```
0xmikko, line 855: could be `unchecked`

0xmikko, line 855: Is it good to put `remainingTokensMask & uint256(-int256(remainingTokensMask))` into `BitMask` librabry and call the method like `.getNextMask()` or `.iter()`?

contracts/credit/CreditFacadeV3.sol
```diff
@@ -358,22 +357,37 @@ contract CreditFacadeV3 is ICreditFacadeV3, ACLNonReentrantTrait {
...
-            FullCheckParams memory fullCheckParams = _multicall(
-                creditAccount, calls, collateralDebtData.enabledTokensMask, CLOSE_CREDIT_ACCOUNT_FLAGS, skipCalls
+            uint256 remainingTokensMask =
+                collateralDebtData.enabledTokensMask.disable(tokensToTransferMask).disable(UNDERLYING_TOKEN_MASK);
+            BalanceWithMask[] memory remainingBalances = BalancesLogic.storeBalances({
+                creditAccount: creditAccount,
+                tokensMask: remainingTokensMask,
+                getTokenByMaskFn: _getTokenByMask
+            });
+
```
0xmikko, line 349: It seems disabling bots in liquidation is obsolete if we keep an account with zeroDebt.

contracts/credit/CreditFacadeV3.sol
```diff
@@ -358,22 +357,37 @@ contract CreditFacadeV3 is ICreditFacadeV3, ACLNonReentrantTrait {
...
             ); // U:[FA-16]
-            collateralDebtData.enabledTokensMask = fullCheckParams.enabledTokensMaskAfter; // U:[FA-16]
+
+            bool success = BalancesLogic.compareBalances({
+                creditAccount: creditAccount,
+                tokensMask: remainingTokensMask,
+                balances: remainingBalances,
+                greater: false
+            });
+
```
0xmikko, line 360: `reportedLoss !=0 ` ?

contracts/credit/CreditFacadeV3.sol
```diff
@@ -358,22 +357,37 @@ contract CreditFacadeV3 is ICreditFacadeV3, ACLNonReentrantTrait {
...
-            closureAction: closeAction,
             collateralDebtData: collateralDebtData,
-            payer: msg.sender,
             to: to,
-            skipTokensMask: skipTokenMask,
-            convertToETH: convertToETH
+            tokensToTransferMask: tokensToTransferMask,
+            convertToETH: convertToETH,
+            isExpired: isExpired
         }); // U:[FA-16]
```
0xmikko, line 373: Should we care about WETH->ETH conversation for liquidators?

contracts/credit/CreditManagerV3.sol
```diff
@@ -219,120 +213,151 @@ contract CreditManagerV3 is ICreditManagerV3, SanityCheckTrait, ReentrancyGuardT
...
         }
 
         _batchTokensTransfer({
             creditAccount: creditAccount,
             to: to,
             convertToETH: convertToETH,
-            tokensToTransferMask: collateralDebtData.enabledTokensMask.disable(skipTokensMask)
-        }); // U:[CM-8, 9]
+            tokensToTransferMask: tokensToTransferMask
+        }); // U:[CM-8]
```
0xmikko, line 358: Check if it makes a tranfer when underlyingBalance == 1?

contracts/credit/CreditManagerV3.sol
```diff
@@ -219,120 +213,151 @@ contract CreditManagerV3 is ICreditManagerV3, SanityCheckTrait, ReentrancyGuardT
...
+                if (amountToLiquidator != 0) {
+                    _safeTokenTransfer({
+                        creditAccount: creditAccount,
+                        token: underlying,
+                        to: to,
+                        amount: amountToLiquidator,
+                        convertToETH: convertToETH
+                    }); // U:[CM-8]
+                    remainingFunds -= amountToLiquidator; // U:[CM-8]
+                }
```
0xmikko, line 348: Pre-requisite: it could set amountToLiquidator =1.

contracts/credit/CreditManagerV3.sol
```diff
@@ -219,120 +213,151 @@ contract CreditManagerV3 is ICreditManagerV3, SanityCheckTrait, ReentrancyGuardT
...
-                to: borrower,
-                amount: remainingFunds,
-                convertToETH: false
-            }); // U:[CM-8]
+        if (tokensToTransferMask & UNDERLYING_TOKEN_MASK != 0) {
+            tokensToTransferMask = tokensToTransferMask.disable(UNDERLYING_TOKEN_MASK);
+
+            unchecked {
+                uint256 amountToLiquidator = remainingFunds - minRemainingFunds;
+                if (amountToLiquidator > underlyingBalance) amountToLiquidator = underlyingBalance;
```
0xmikko, line 337: This code block is obsolete if `loss ! =0`

0xmikko (commented):
A few comments:

1. It's better to use enum `Comparison.GREATER_THAN_SAVED` of shorter `Comparison.GREATER` than boolean flag.
2. Didn't find withdrawIfZeroDebt function there, alternatively it could be added here: 
```
 if (IWithdrawalManagerV3(withdrawalManager).delay() == 0) {
            address borrower = getBorrowerOrRevert({creditAccount: creditAccount});
            _safeTokenTransfer({
                creditAccount: creditAccount,
                token: token,
                to: borrower,
                amount: amount,
                convertToETH: false
            }); // U:[CM-27]
        } else {
 ```
additinally checking that debt == 0.

**Review by 0xmikko — COMMENTED** [review]
A few comments:

1. It's better to use enum `Comparison.GREATER_THAN_SAVED` of shorter `Comparison.GREATER` than boolean flag.
2. Didn't find withdrawIfZeroDebt function there, alternatively it could be added here: 
```
 if (IWithdrawalManagerV3(withdrawalManager).delay() == 0) {
            address borrower = getBorrowerOrRevert({creditAccount: creditAccount});
            _safeTokenTransfer({
                creditAccount: creditAccount,
                token: token,
                to: borrower,
                amount: amount,
                convertToETH: false
            }); // U:[CM-27]
        } else {
 ```
additinally checking that debt == 0.

contracts/credit/CreditFacadeV3.sol
lekhovitsky: ✅

PR #155, opened 2023-11-09
feat: minor credit facade improvements

lekhovitsky:
In this PR:
* slippage check is slightly refined: now, instead of a single `revertIfReceivedLessThan` call, a user can place multiple `storeExpectedBalances` and `compareBalances` calls in the multicall
* full check params sanitization is moved from credit manager to credit facade
* `withdrawCollateral` call is now allowed in `openCreditAccount`
* other minor fixes and improvements


0xmikko (approved):

contracts/libraries/CollateralLogic.sol
```diff
@@ -121,9 +121,6 @@ library CollateralLogic {
 
         for (uint256 i; i < len;) {
             address token = quotedTokens[i]; // U:[CLL-4]
-            // `quotedTokens` has fixed length (`maxEnabledTokens`), but not all of its entries are actual tokens;
-            // encountering `address(0)` for the first time indicates that all quoted tokens have been processed
-            if (token == address(0)) break; // U:[CLL-4]
```
lekhovitsky, line 126: This is no longer needed as `quotedTokens` always has correct length

contracts/libraries/QuotasLogic.sol
```diff
@@ -32,7 +32,7 @@ library QuotasLogic {
         returns (uint128)
     {
         // `quoted` is `uint96`, and `cumulativeIndex / RAY` won't reach `2 ** 32` in reasonable time, so casting is safe
-        return uint128(uint256(quoted) * (cumulativeIndexNow - cumulativeIndexLU) / RAY); // U:[QL-2]
+        return quoted <= 1 ? 0 : uint128(uint256(quoted) * (cumulativeIndexNow - cumulativeIndexLU) / RAY); // U:[QL-2]
```
lekhovitsky, line 35: No need to perform any calculations if there is no quota (especially useful for `quoted == 1` case).

contracts/pool/PoolQuotaKeeperV3.sol
```diff
@@ -375,7 +374,7 @@ contract PoolQuotaKeeperV3 is IPoolQuotaKeeperV3, ACLNonReentrantTrait, Contract
 
         // The rate will be set during a general epoch update in the gauge
         quotaTokensSet.add(token); // U:[PQK-5]
-        totalQuotaParams[token].cumulativeIndexLU = uint192(RAY); // U:[PQK-5]
+        totalQuotaParams[token].cumulativeIndexLU = 1; // U:[PQK-5]
```
lekhovitsky, line 377: It is an artifact from times when quota index was multiplicative. With additive index, any non-zero number would work (zero indicates that token is not quoted)

contracts/credit/CreditManagerV3.sol
```diff
@@ -913,8 +901,8 @@ contract CreditManagerV3 is ICreditManagerV3, SanityCheckTrait, ReentrancyGuardT
         returns (uint256 tokensToEnable, uint256 tokensToDisable)
     {
         CreditAccountInfo storage currentCreditAccountInfo = creditAccountInfo[creditAccount];
-        if (quotaChange > 0 && currentCreditAccountInfo.debt == 0) {
-            revert IncreaseQuotaOnZeroDebtAccountException();
+        if (currentCreditAccountInfo.debt == 0) {
+            revert UpdateQuotaOnZeroDebtAccountException();
```
lekhovitsky, line 905: In theory that doesn't change anything since zero-debt account must have no quotas anyways, but it looks more robust this way.

contracts/credit/CreditManagerV3.sol
```diff
@@ -420,6 +415,12 @@ contract CreditManagerV3 is ICreditManagerV3, SanityCheckTrait, ReentrancyGuardT
             }
         }
 
+        // zero-debt is a special state that disables collateral checks so having quotas on
+        // the account should be forbidden as they entail debt in a form of quota interest
+        if (newDebt == 0 && collateralDebtData.quotedTokens.length != 0) {
+            revert DebtToZeroWithActiveQuotasException();
+        }
```
lekhovitsky, line 422: Doing this check in all cases is a bit more robust

contracts/credit/CreditFacadeV3.sol
```diff
@@ -589,6 +595,16 @@ contract CreditFacadeV3 is ICreditFacadeV3, ACLNonReentrantTrait {
...
+
+                        if (fullCheckParams.minHealthFactor < PERCENTAGE_FACTOR) {
+                            revert CustomHealthFactorTooLowException(); // U:[FA-24]
+                        }
+
+                        uint256 hintsLen = fullCheckParams.collateralHints.length;
+                        for (uint256 j; j < hintsLen; ++j) {
+                            uint256 mask = fullCheckParams.collateralHints[j];
+                            if (mask == 0 || mask & mask - 1 != 0) revert InvalidCollateralHintException(); // U:[FA-24]
+                        }
```
lekhovitsky, line 607: Since full check params are set here, it makes sense to validate them here as well

contracts/credit/CreditFacadeV3.sol
```diff
@@ -43,8 +43,7 @@ import {PERCENTAGE_FACTOR} from "@gearbox-protocol/core-v2/contracts/libraries/C
 // EXCEPTIONS
 import "../interfaces/IExceptions.sol";
 
-uint256 constant OPEN_CREDIT_ACCOUNT_FLAGS =
-    ALL_PERMISSIONS & ~(DECREASE_DEBT_PERMISSION | WITHDRAW_COLLATERAL_PERMISSION);
+uint256 constant OPEN_CREDIT_ACCOUNT_FLAGS = ALL_PERMISSIONS & ~DECREASE_DEBT_PERMISSION;
```
lekhovitsky, line 46: Doesn't change security assumptions since once could simply do `openCreditAccount` with empty `calls` followed by `multicall` in one transaction to achieve the same result

contracts/interfaces/IExceptions.sol
```diff
@@ -184,15 +178,18 @@ error BorrowedBlockLimitException();
...
 
-/// @notice Thrown if expected balances are attempted to be set through `revertIfReceivedLessThan` twice
+/// @notice Thrown if expected balances are attempted to be set twice without performing a slippage check
 error ExpectedBalancesAlreadySetException();
 
+/// @notice Thrown if attempting to perform a slippage check when excepted balances are not set
+error ExpectedBalancesNotSetException();
+
+/// @notice Thrown if balance of at least one token is less than expected during a slippage check
+error BalanceLessThanExpectedException();
```
3sGgpQ8H, line 191: This error could be made more useful by adding certain parameters into it.

contracts/interfaces/IExceptions.sol
```diff
@@ -217,6 +214,12 @@ error TooManyApprovedBotsException();
 /// @notice Thrown when attempting to give a bot unexpected permissions
 error UnexpectedPermissionsException();
 
+/// @notice Thrown when a custom HF parameter lower than 10000 is passed into the full collateral check
```
3sGgpQ8H, line 217: Using exact numbers in comments is a bad practice, as once such number change, it would be hard to fix all the relevant comments.  Consider changing `10000` with `PERCENTAGE_FACTOR` or other appropriate constant name.

contracts/interfaces/IExceptions.sol
```diff
@@ -217,6 +214,12 @@ error TooManyApprovedBotsException();
...
+error CustomHealthFactorTooLowException();
```
3sGgpQ8H, line 218: This error could be made more useful by adding certain parameters into it.

contracts/interfaces/IExceptions.sol
```diff
@@ -217,6 +214,12 @@ error TooManyApprovedBotsException();
...
+
+/// @notice Thrown when submitted collateral hint is not a valid token mask
+error InvalidCollateralHintException();
```
3sGgpQ8H, line 221: This error could be made more useful by adding certain parameters into it.

PR #166, opened 2023-12-06
fix: quota min value fixed to proper formula

0xmikko:
Change: fix previous mistake, when during collateral computation, it uses formula: 
` weightedValueUSD = Math.min(valueUSD, quotaUSD) * liquidationThreshold / PERCENTAGE_FACTOR;`

This approach is too risky: in case of downgrading LT, many users who have enough collateral in quoted token (for example x2 of debt size) could be liquidated - they could not take extra quotas from PQK if limit is reached. 

So, in this case, lowering LT could be a problem because many users who have high hf could be liquidated without any reason.

### DIFF
--- a/contracts/libraries/CollateralLogic.sol
+++ b/contracts/libraries/CollateralLogic.sol
@@ -263,7 +263,7 @@ library CollateralLogic {
             unchecked {
                 valueUSD = convertToUSDFn(priceOracle, balance - 1, token); // U:[CLL-1]
             }
-            weightedValueUSD = Math.min(valueUSD, quotaUSD) * liquidationThreshold / PERCENTAGE_FACTOR; // U:[CLL-1]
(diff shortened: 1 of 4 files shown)


lekhovitsky (approved):

contracts/libraries/CollateralLogic.sol
```diff
@@ -263,7 +263,7 @@ library CollateralLogic {
             unchecked {
                 valueUSD = convertToUSDFn(priceOracle, balance - 1, token); // U:[CLL-1]
             }
-            weightedValueUSD = Math.min(valueUSD, quotaUSD) * liquidationThreshold / PERCENTAGE_FACTOR; // U:[CLL-1]
+            weightedValueUSD = Math.min(valueUSD * liquidationThreshold / PERCENTAGE_FACTOR, quotaUSD); // U:[CLL-1]
```
3sGgpQ8H, line 266: Consider using the mulDiv function or preventing phantom overflow in some other way.

PR #171, opened 2024-01-15
feat: partial liquidation bot

Van0k:


contracts/bots/PartialLiquidationBot.sol
```diff
@@ -0,0 +1,179 @@
...
+            /// minimal debt limit is violated as a result
+            (uint256 minDebt,) = ICreditFacadeV3(creditFacade).debtLimits();
+            if (amountInUnderlying > CreditLogic.calcTotalDebt(cdd) - minDebt) {
+                revert CantPartialLiquidateBelowMinDebt();
+            }
+        }
+
+        if (amountInUnderlying > params.maxAmountInUnderlying) revert AmountUnderlyingLargerThanMax();
+
+        IERC20(underlying).transferFrom(msg.sender, address(this), amountInUnderlying);
```
Van0k: Change to transfer directly to Credit Account + enable underlying

contracts/bots/PartialLiquidationBot.sol
```diff
@@ -0,0 +1,179 @@
...
+
+        /// Since we are computing debt and collateral before liquidation,
+        /// we need to update prices beforehand, if needed
+        _applyPriceUpdates(params.priceUpdates, priceOracle);
+
+        CollateralDebtData memory cdd = ICreditManagerV3(params.creditManager).calcDebtAndCollateral(
+            params.creditAccount, CollateralCalcTask.DEBT_COLLATERAL
+        );
+
+        if (cdd.twvUSD >= cdd.totalDebtUSD) revert CreditAccountNotLiquidatableException();
```
Van0k: Replace with isLiquidatable

contracts/bots/PartialLiquidationBot.sol
```diff
@@ -0,0 +1,158 @@
...
+
+        address creditFacade = ICreditManagerV3(params.creditManager).creditFacade();
+
+        // It's techically possible to pass underlying as assetOut
+        // However, this would reduce the HF of the account, which will result
```
lekhovitsky: does it actually hold if `repay` is `true`? something like (i) buy underlying at discount (which decreases hf and (ii) decrease debt (which increases hf and cancels the previous effect)

contracts/bots/PartialLiquidationBot.sol
```diff
@@ -0,0 +1,158 @@
...
+        // There is a limit imposed on the maximal amount of underlying that the liquidator can swap
+        // If the debt is repaid, then that limit is totalDebt - minDebt - underlyingBalance (i.e., exactly enough to reduce debt to minDebt)
+        // If the debt is not repaid, then that limit is totalDebt - underlyingBalance (i.e., exactly enough to fully liquidate the user in the future)
+
+        uint256 amountInUnderlying;
+        {
+            uint256 underlyingBalance = IERC20(underlying).balanceOf(params.creditAccount);
+            (uint256 minDebt,) = ICreditFacadeV3(creditFacade).debtLimits();
+
+            uint256 repayable = CreditLogic.calcTotalDebt(cdd) - (params.repay ? minDebt : 0);
```
lekhovitsky: limiting the amount of underlying added by liquidator to `assetOut`'s quota seems a bit more reasonable to prevent liquidator from taking disabled tokens or tokens that didn't count towards collateral in general

it's a bit unfortunate though that we don't enforce all non-underlying tokens to be quoted.

contracts/bots/PartialLiquidationBot.sol
lekhovitsky: seems more robust to explicitly forbid setting `assetOut` to underlying in order not to create the money printer

contracts/bots/PartialLiquidationBot.sol
```diff
@@ -0,0 +1,158 @@
...
+
+            if (underlyingBalance >= repayable) revert NothingToLiquidateException();
```
lekhovitsky: in this case we rely on full liquidations, right?

contracts/bots/PartialLiquidationBot.sol
Van0k: Fair point, I'll add a check

contracts/bots/PartialLiquidationBot.sol
Van0k: Yes, it's sufficient to swap enough underlying to reach totalDebt, and then you can fully liquidate cheaply (repay is separate, since you cannot go below minDebt). One caveat is that if I do a partial liquidation get totalDebt of underlying on the account, and then try to fully liquidate on the next block, it will likely not be enough. It makes sense to add a small buffer to account for that

contracts/bots/PartialLiquidationBot.sol
Van0k: Fair point, will fix

contracts/bots/PartialLiquidationBot.sol
```diff
@@ -0,0 +1,280 @@
...
+    ///         Accepts the amount of assetOut that the liquidator wishes to receive
+    /// @param creditManager Credit Manager where the account resides
+    /// @param creditAccount Credit Account to liquidate
+    /// @param assetOut Asset to receive for underlying
+    /// @param amountOut Amount of assetOut to receive. Note that depending on the CA's balance in assetOut and
+    ///                 the maximal swappable amount of underlying, the actual received amount may be less
+    ///                 (with the spent amount adjusted accordingly)
+    /// @param repay Whether to repay debt or only perform a swap
+    /// @param priceUpdates Data for updatable price feeds, if any are required to compute account health
+    function partialLiquidateExactOut(
```
lekhovitsky, line 80: Looks like we can combine `partialLiquidateExactIn` and `partialLiquidateExactOut` and only add a single flag `exactIn` since the code is way too similar to keep two functions

contracts/bots/PartialLiquidationBot.sol
```diff
@@ -0,0 +1,280 @@
...
+        IERC20(params.underlying).transferFrom(msg.sender, params.creditAccount, params.amountIn);
+
+        ICreditFacadeV3(params.creditFacade).botMulticall(params.creditAccount, _getMultiCall(params));
+
+        CollateralDebtData memory cdd = ICreditManagerV3(params.creditManager).calcDebtAndCollateral(
+            params.creditAccount, CollateralCalcTask.DEBT_COLLATERAL
+        );
+
+        if (cdd.twvUSD * PERCENTAGE_FACTOR < cdd.totalDebtUSD * THRESHOLD_HEALTH_FACTOR) {
+            revert HealthFactorTooLowException();
```
lekhovitsky: Maybe we can pass `minHealthFactor` to full collateral check and then repeat the calculation if `creditManager.version() == 3_00`? would save considerable amounts of gas for all new credit managers

contracts/bots/PartialLiquidationBot.sol
Van0k: Unfortunately, the function ends up having too many params and causes a "stack too deep" error

contracts/bots/PartialLiquidationBot.sol
Van0k: Fixed

contracts/bots/PartialLiquidationBot.sol
```diff
@@ -0,0 +1,292 @@
...
+        params.assetOut = assetOut;
+        params.amountIn = amountIn;
+        params.amountOut = amountOut;
+        params.exactIn = exactIn;
+        params.repay = repay;
+        params.to = to;
+        params.creditFacade = ICreditManagerV3(params.creditManager).creditFacade();
+        params.cmVersion = ICreditManagerV3(params.creditManager).version();
+        params.underlying = ICreditManagerV3(params.creditManager).underlying();
+        params.priceOracle = ICreditManagerV3(creditManager).priceOracle();
```
0xmikko, line 125: What the reason to make `priceOracle` call here, if there is no evidence that applyPriceUpdates will be used. Could be called from the function in lazy manner as in CreditFacade

contracts/bots/PartialLiquidationBot.sol
```diff
@@ -0,0 +1,292 @@
...
+            calls[2] = MultiCall({
+                target: params.creditFacade,
+                callData: abi.encodeCall(
+                    ICreditFacadeV3Multicall.decreaseDebt, (IERC20(params.underlying).balanceOf(params.creditAccount))
+                    )
+            });
+        }
+
+        if (params.cmVersion > 3_00) {
+            calls[4] = MultiCall({
```
0xmikko: Why '4' is hardcoded here? At first glance, it's out of bounds (3 is max if length equals 4). What if repay is `false` and cmVersion > 3_00? It seems, it should be computed like `calls.length-1` or by formula similar in L240

contracts/bots/PartialLiquidationBot.sol
```diff
@@ -0,0 +1,292 @@
...
+        LiquidationParams memory params =
+            _prepareParams(creditManager, creditAccount, assetOut, 0, amountOut, to, repay, false);
+
+        /// Since we are computing debt and collateral before liquidation,
+        /// we need to update prices beforehand, if needed
+        _applyPriceUpdates(priceUpdates, params.priceOracle);
+
+        _retrieveAndCheckDebt(params);
+
+        _liquidate(params);
```
0xmikko, line 98: It seems, that repay could be a parameter here

contracts/bots/PartialLiquidationBot.sol
```diff
@@ -0,0 +1,292 @@
...
+    ) internal view returns (LiquidationParams memory params) {
+        params.creditManager = creditManager;
+        params.creditAccount = creditAccount;
+        params.assetOut = assetOut;
+        params.amountIn = amountIn;
+        params.amountOut = amountOut;
+        params.exactIn = exactIn;
+        params.repay = repay;
+        params.to = to;
+        params.creditFacade = ICreditManagerV3(params.creditManager).creditFacade();
```
0xmikko: Why it's here params.creditManager instead `creditManager` itself

contracts/bots/PartialLiquidationBot.sol
```diff
@@ -0,0 +1,292 @@
...
+
+            maxAmountIn = repayable - underlyingBalance;
+            maxAmountOut = IERC20(params.assetOut).balanceOf(params.creditAccount);
+        }
+
+        if (params.exactIn) {
+            (params.amountIn, params.amountOut) = _getAmountsExactIn(params, maxAmountIn, maxAmountOut);
+        } else {
+            params.amountOut = params.amountOut == type(uint256).max
+                ? IERC20(params.assetOut).balanceOf(params.creditAccount)
```
0xmikko: Could we just use `maxAmountOut` here instead one more call?

0xmikko (commented):
Is it possible to rewrite without memory structure without `stack to deep` ?

contracts/bots/PartialLiquidationBot.sol
Van0k: It's needed regardless of price updates for exchange rate computation

contracts/bots/PartialLiquidationBot.sol
Van0k: Fixed

contracts/bots/PartialLiquidationBot.sol
Van0k: "repay" is already in params

contracts/bots/PartialLiquidationBot.sol
Van0k: fixed

contracts/bots/PartialLiquidationBot.sol
Van0k: fixed

contracts/bots/PartialLiquidationBot.sol
0xmikko: I understood that it's used in this function only, proboaly, it's better to move it out from params and use directly here.

contracts/bots/PartialLiquidationBot.sol
0xmikko: Is it correct statement that `if (underlyingBalance >= repayable) revert NothingToLiquidateException()`?
Debt is 1000 USDC, 1000 USDC. LTunderlying = 0.955, so hf = 0.955 and account should be liquidated

0xmikko: In previous statement, partial liquidation could be for another token with lower LT, which could cover the gap, but this contract will revert.

contracts/bots/PartialLiquidationBot.sol
lekhovitsky: One way to overcome it is to combine `amountIn` and `amountOut` args of `_prepareParams` to just `amount` (the other is always zero) and dispatch there based on `exactIn` flag:

```solidity
LiquidationParams memory params = 
    _prepareParams(creditManager, creditAccount, assetOut, amount, to, repay, exactIn);
```

And then in `_prepareParams`
```solidity
(params.amountIn, params.amountOut) = exactIn ? (amount, 0) : (0, amount);
```

contracts/bots/PartialLiquidationBot.sol
Van0k: If underlyingBalance >= repayable (i.e., totalDebt), then the account can be just fully liquidated by calling ICreditFacadeV3.liquidateCreditAccount

lekhovitsky: can rename the exception to indicate that partial liquidation is not needed in this case

contracts/bots/PartialLiquidationBot.sol
Van0k: It is also used in _getMulticall()

contracts/bots/PartialLiquidationBot.sol
0xmikko: It seems incorrect. Let's consider the case:

totalDebt: 1000 USC
Balances:
[USDC, underlying] : 1000 USDC
[token A] : 100 Atoken

100 Atoken represent the ```value = totalDebt * (1-LTunderlying) / LTa``` So, last block it was:

```totalDebt == 1000 USDC * LTunderlying + value of token A * LTa``` 

Then price of token A was dropped for 1, as result: ```totalDebt > twv```, so ```HF < 1``` => account is liquidatable. 

Liquidator should pay 1000 USDC back to pool and then swap 100 Atoken for liquidation fee + his profits. However, liquidity of token A is not enough...

contracts/bots/PartialLiquidationBot.sol
Van0k: Tried this and a few other ways, still hitting the stack limit for me

PR #199, opened 2024-04-18
feat: `BotListV3` improvements

lekhovitsky:
In this PR:
* new `IBotV3` interface that defines bot's required permissions which are checked in the bot list
* special permission bots are deprecated
* `BotListV3` interface changes:
  * redundant `creditManager` parameter removed from most functions that operate on `creditAccount`
  * `EraseBot` event removed in favor of `SetBotPermissions(..., 0)`
* minor implementation improvements that make code more consistent
* increase `BotListV3` version to `3_10`


lekhovitsky:
Worth noting that requiring bots to conform to the interface means that (i) users won't be able to "share ownership" of their credit account by setting some SC account as bot and (ii) DAO won't be able to set EOA as special bot (somehow this slipped into the original bot list unnoticed)

UPD: the latter doesn't matter since special bots are removed after all

PR #202, opened 2024-05-01
feat: treat all non-underlying tokens as quoted

lekhovitsky:
All collateral tokens are now considered quoted (credit configurator forbids adding non-quoted tokens as collateral), while underlying token is always considered enabled. The former allows to make the system stricter and remove a huge pile of dead code related to enabling/disabling tokens, while the latter allows to prevent liquidations during pause by sending underlying directly to credit account.

While it's encouraged to use new credit managers going forward, new facade and configurator are compatible with existing ones, with two important remarks:
* all non-quoted collateral tokens must be made such **before** upgrading credit facades and configurators;
* open credit accounts with disabled underlying will only have it enabled after the first multicall. Before that, sending underlying directly won't prevent liquidation and might result in the loss of funds.

Additionally, this PR removes all functionality related to `revokeAdapterAllowances`.


lekhovitsky:
Note on compatibility of new facades with old credit managers: 
* New facades enable underlying token after every multicall regardless of its balance.
* Old managers disable zero-balance tokens in full collateral check unless the target TWV is reached. But, since underlying is now the only non-quoted token, then a successful full collateral check either stops before checking underlying's balance, or sees a non-zero underlying balance, i.e., it never disables underlying.

PR #243, opened 2024-07-16
feat: deployment simplification

lekhovitsky:
In this PR:
* `PoolQuotaKeeperV3` is now deployed inside `PoolV3`'s constructor, which allows to store it as immutable variable making interactions with the system cheaper
* All the contracts that can read `acl` and `contractsRegister` instead of accepting a constructor argument now do so, reducing the chances of misconfiguration
* Constructors now consistently revert with `ZeroAddressException` 
* Better validation in `USDT_Transfer` trait
* Silence warnings in tests

### DIFF
--- a/contracts/credit/CreditConfiguratorV3.sol
+++ b/contracts/credit/CreditConfiguratorV3.sol
@@ -21,6 +21,7 @@ import {CreditManagerV3} from "./CreditManagerV3.sol";
 import {ICreditConfiguratorV3, AllowanceAction} from "../interfaces/ICreditConfiguratorV3.sol";
 import {IPoolQuotaKeeperV3} from "../interfaces/IPoolQuotaKeeperV3.sol";
 import {IPriceOracleV3} from "../interfaces/IPriceOracleV3.sol";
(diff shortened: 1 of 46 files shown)


contracts/credit/CreditConfiguratorV3.sol
```diff
@@ -62,10 +63,9 @@ contract CreditConfiguratorV3 is ICreditConfiguratorV3, ControlledTrait, Reentra
     }
 
     /// @notice Constructor
-    /// @param _acl ACL contract address
     /// @param _creditManager Credit manager to connect to
     /// @dev Copies allowed adaprters from the currently connected configurator
-    constructor(address _acl, address _creditManager) ControlledTrait(_acl) {
+    constructor(address _creditManager) ControlledTrait(IPoolV3(CreditManagerV3(_creditManager).pool()).acl()) {
```
0xmikko: it's better to add sanity check for _creditManager is not zero address. We implemeted it in prev versions as:
_creditManager == address(0) ? address(0) : IPoolV3....

0xmikko (approved):

PR #246, opened 2024-08-14
feat: `maxEnabledTokens` is now immutable 

lekhovitsky:
Fixes https://github.com/spearbit-audits/review-gearbox/issues/89


StErMi:
As far as I see, the code seems to be ok and there should be no side effects considering that `CreditManager` is a separate entity and existing credit managers cannot be upgraded.

@lekhovitsky would you mind giving more context on the decision to make it `immutable` instead of directly fixing the current behavior?

Please be aware that because `CreditManager`s are immutable (can't be upgraded), existing contracts are vulnerable to this issue. This is important information to be aware of and document internally.

lekhovitsky:
@StErMi We've never encountered the need to re-configure `maxEnabledTokens` in practice. On the other hand, we're able to enforce some additional security checks when it's immutable (as can be seen in https://github.com/Gearbox-protocol/core-v3/pull/261, on which I'll elaborate as we get to it).

PR #248, opened 2024-08-14
fix: add comments that disabled tokens can be seized by liquidator

lekhovitsky:
Acknowledges:
* https://github.com/spearbit-audits/review-gearbox/issues/70
* https://github.com/spearbit-audits/review-gearbox/issues/79
* https://github.com/spearbit-audits/review-gearbox/issues/81
* https://github.com/spearbit-audits/review-gearbox/issues/95


StErMi:
@lekhovitsky now that I think about it, should a user be able to perform a "quota increase" for a token that has `LT=0`? 

Purchasing such an amount of quota (which has a fixed fee + quota interest) does not contribute at all to the increase in user's borrowing power or HF. In reality, we have the opposite effect because, as mentioned above, you are increasing your debt.

There's no check about the token's liquidation threshold in the `CreditFacade`, `CreditManager` or `PoolQuotaKeeper` and as far as I see there is no direct connection between having a token (which can be configured as a valid collateral with LT = 0) but also being not-purchasable as a quota token (if it has been already enabled as a quota token).

Am I missing something down the line, or is it reasonable to revert the increase in quota for an `LT=0` token?

The only maybe related mechanism would be to have such `LT=0` token also configured as `forbidden` but I think that it has much more unwanted side effects (because of the forbidden logic inside the `CreditFacade`).

lekhovitsky:
@StErMi this will likely be handled by having zero limits for such tokens in the quota keeper. In some cases, however, it might make sense to allow users to enable such tokens by purchasing quota to count towards `totalValue` (and thus towards `remainingFunds`) during liquidations, even though they don't affect ones health factor.

Anyways, since this does not pose a direct security risk (and we don't consider loss of funds due to unintended use of the system to be a huge security risk), this is much easier to deal with on the configuration level instead of imposing additional constraints in the code.

PR #249, opened 2024-08-15
fix: make pool shares transfers pausable

lekhovitsky:
Fixes https://github.com/spearbit-audits/review-gearbox/issues/11


StErMi:
The implementations now reverts the transfer of tokens when the `Pool` is paused by overriding the `ERC20._transfer` function.

A cleaner option would have been to override and implement the same logic in the `_beforeTokenTransfer` hook, but such hook would also have been called in other flows like `ERC20._mint` and `ERC20._burn` which are already covered directly in the `PoolV3` implementation.

lekhovitsky:
`_transfer` was chosen over `_beforeTokenTransfer` exactly to avoid duplicate checks (and make as little changes to the code as possible).

PR #252, opened 2024-08-15
fix: ensure price oracle has price feeds for all collaterals

lekhovitsky:
Fixes https://github.com/spearbit-audits/review-gearbox/issues/71


StErMi:
With the following implementation, we know that both the `CreditConfigurator` and `CreditManager` will be in a correct state when the `priceOracle` is configured.

Now the logic will revert when 

- `priceOracle` is `address(0)`
- the `priceFeed` of the token (underlying or one of the configured collateral token in the list) is `address(0)` (for the primary feed)
- the `priceFeed` of the token (underlying or one of the configured collateral token in the list) for the primary feed cannot provide a valid price

Note that with the current implementation, there the non-safe `priceFeed` (`address(0)` or valid price value) is checked.

The secondary (safe-price) prices are non required for Gearbox, but they could be checked anyway if the Gearbox team consider it an important security measure to be added to the configuration logic of the new `priceFeed`.

lekhovitsky:
The main issue with checking reserve price feeds too is the fact that there might not be an alternative data source for a given token. In this case, it's clearly better to be able to list such token without allowing to withdraw against it compared to not being able to list it at all.

PR #253, opened 2024-08-15
fix: don't skip price check when adding updatable feeds

lekhovitsky:
Fixes https://github.com/spearbit-audits/review-gearbox/issues/21


StErMi:
With the current implementation now `PriceOracleV3` executes the `_checkAnswer` internal function to validate the oracle's answer even if the `priceFeed` is "updatable".

This logic is performed for both the primary and reserve token's `priceFeed`

PR #254, opened 2024-08-15
fix: make bot required permissions check strict

lekhovitsky:
Fixes https://github.com/spearbit-audits/review-gearbox/issues/25


StErMi:
The `BotListV3.setBotPermissions` function now correctly reverts if the user tries to give more permission than the minimum one required by the bot. The documentation and revert reason have been updated accordingly.

PR #255, opened 2024-08-15
feat: `BotListV3` config functions are now one-way

lekhovitsky:
Fixes https://github.com/spearbit-audits/review-gearbox/issues/26

### DIFF
--- a/contracts/core/BotListV3.sol
+++ b/contracts/core/BotListV3.sol
@@ -131,20 +131,20 @@ contract BotListV3 is IBotListV3, SanityCheckTrait, Ownable {
         return _botInfo[bot].forbidden;
     }
 
-    /// @notice Sets `bot`'s status to `forbidden`
-    function setBotForbiddenStatus(address bot, bool forbidden) external override onlyOwner {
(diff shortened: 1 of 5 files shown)


StErMi:
@lekhovitsky a couple of questions/feedback

The `BotListV3.eraseAllBotPermissions` called during `CreditFacadeV3.closeCreditAccount` and `BotListV3.setBotPermissions` will not revert because you have disabled the possibility to "blacklist" a `CreditManager` once it has been whitelisted to operate on the `BotListV3`. 

This resolves the issues reported, but it also removes the feature to disallow a `CreditManager` to interact with the `BotListV3` which was available before such changes. Would you mind explaining the motivation of this decision and why it should not create a possible issue in the future?

The same "loss" of a feature happens for the "forbid bot" that can't be re-activated again, but I find it less problematic because someone could re-deploy a bot and whitelist it. It's more of a hassle than a real problem.

lekhovitsky:
@StErMi there are no practical reasons to forbid credit managers in the bot list as even a malicious one can't mess with credit accounts not associated with it.

PR #260, opened 2024-08-16
fix: reset credit account variables upon opening and closure

lekhovitsky:
Fixes:
* https://github.com/spearbit-audits/review-gearbox/issues/40
* https://github.com/spearbit-audits/review-gearbox/issues/41

### DIFF
--- a/contracts/credit/CreditManagerV3.sol
+++ b/contracts/credit/CreditManagerV3.sol
@@ -216,6 +216,8 @@ contract CreditManagerV3 is ICreditManagerV3, SanityCheckTrait, ReentrancyGuardT
             sstore(slot, 1)
         } // U:[CM-6]
 
+        newCreditAccountInfo.enabledTokensMask = UNDERLYING_TOKEN_MASK; // U:[CM-6]
+
         creditAccountsSet.add(creditAccount); // U:[CM-6]
     }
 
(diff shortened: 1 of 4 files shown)


StErMi:
The PR addresses the linked issues

contracts/credit/CreditManagerV3.sol
```diff
@@ -673,7 +691,8 @@ contract CreditManagerV3 is ICreditManagerV3, SanityCheckTrait, ReentrancyGuardT
         CreditAccountInfo storage currentCreditAccountInfo = creditAccountInfo[creditAccount];
 
         cdd.debt = currentCreditAccountInfo.debt; // U:[CM-20]
-        cdd.cumulativeIndexLastUpdate = currentCreditAccountInfo.cumulativeIndexLastUpdate; // U:[CM-20]
```
StErMi, line 676: @lekhovitsky would you mind explaining the reason for this change? As far as I see, it has not been requested in the linked issues recommendations.

StErMi: One discrepancy that I see is that the value retrieved by querying directly `creditAccountInfo[creditAccount].cumulativeIndexLastUpdate` will be different from the one returned by `CreditManager.calcDebtAndCollateral(creditAccount, task)` (when `task !== FULL_COLLATERAL_CHECK_LAZY`)

The first one would return a `cumulativeIndexLastUpdate > 0` (the value that is directly stored in the blockchain), the second one instead will return `0`.

I mean it seems that there's no harm because as you said in the comment the **interest** index (and last update) will not be used by the current codebase inside the calculation when `debt == 0` (just because you always early return when there's no debt) but I was curious to know which is the reason for the change and if you think that this divergence could be an issue for external sources that rely on this data even if we are in an edge case (`debt == 0`)

lekhovitsky: `creditAccountInfo` is just a raw getter that is not internal for ~~arbitrary~~ historic reasons. The code that uses it shall bear all responsibility for handling edge cases.

Regarding the change: `cumulativeIndexLastUpdate` is always in the denominator, and using it when `debt == 0` is so wrong that we'd prefer this division to panic to avoid unexpected behaviors.

PR #261, opened 2024-08-16
feat: add checks on debt limits

lekhovitsky:
Fixes https://github.com/spearbit-audits/review-gearbox/issues/56

Additionally, introduces a crucial check that maxDebt / minDebt doesn't exceed a safety threshold.

### DIFF
--- a/contracts/credit/CreditConfiguratorV3.sol
+++ b/contracts/credit/CreditConfiguratorV3.sol
@@ -442,6 +442,7 @@ contract CreditConfiguratorV3 is ICreditConfiguratorV3, ACLNonReentrantTrait {
 
     /// @notice Sets the new price oracle contract in the credit manager
     /// @param newPriceOracle New price oracle
+    /// @dev Reverts if USD value of the current min debt is zero according to `newPriceOracle`
     /// @dev Checks that `newPriceOracle` has prices for all collateral tokens
     function setPriceOracle(address newPriceOracle)
         external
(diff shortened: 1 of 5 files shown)


StErMi:
@lekhovitsky would you mind explaining this new check?

```solidity
        if (maxDebt * CreditManagerV3(creditManager).maxEnabledTokens() > minDebt * 100) {
            revert IncorrectLimitsException(); // I:[CC-15]
        }
```

The above check is out of the scope of the fix review and I don't see a direct correlation between the `minDebt` or `maxDebt` with the number of max enabled token that are related to the collateral enabled (some quota has been purchased to later on borrow)

StErMi:
Excluding the comment above, the implementation seems correct. 

Note that the above checks are valid **only** for the current (when the checks are performed) configuration of the `PriceOracleV3` instance.

If `PriceOracleV3.setPriceFeed(underlying, ...)` is called or if the value returned by `PriceOracleV3.convertToUSD(minDebt, underlying)` is later on equal to zero, those checks are not valid anymore. For this reason, they cannot be considered an invariant that is always held after it has been proven at the moment of execution of `CreditConfiguratorV3.setPriceOracle` or  `CreditConfiguratorV3.setDebtLimits`

Note also that these checks are implemented directly on the `CreditConfiguratorV3` instance but the `CreditManagerV3` is deployed with the `creditConfigurator == msg.sender` and there's no assurance that the deployer is indeed a valid credit configurator instead of an EOA or another contract that won't perform those checks.

As already mentioned in other findings, Gearbox should consider moving to the base contract (in this case the `CreditManager`) all these checks.

StErMi:
Note also that `_setLimits` (where some of the checks on the `minDebt` value are made) is also indirectly called when the `CreditConfiguratorV3.setCreditFacade(address newCreditFacade, bool migrateParams)` is executed with `migrateParams == true`.

Such transaction could revert if the current USD conversion of the current `minDebt` value is equal to zero. This is a minor issue because the configurator can work around the situation by updating the `minDebt` value and re-execute the `setCreditFacade` function.

contracts/credit/CreditConfiguratorV3.sol
```diff
@@ -450,9 +451,12 @@ contract CreditConfiguratorV3 is ICreditConfiguratorV3, ACLNonReentrantTrait {
     {
         if (newPriceOracle == CreditManagerV3(creditManager).priceOracle()) return;
 
+        (uint128 minDebt,) = CreditFacadeV3(creditFacade()).debtLimits();
+        if (IPriceOracleV3(newPriceOracle).convertToUSD(minDebt, underlying) == 0) revert IncorrectPriceException(); // I:[CC-21]
+
         uint256 num = CreditManagerV3(creditManager).collateralTokensCount();
         unchecked {
-            for (uint256 i; i < num; ++i) {
+            for (uint256 i = 1; i < num; ++i) {
```
cryptarasecurity, line 459: Skipping the underlaying `getPrice` check. The underlaying oracle is separetly checked via `convertToUSD` and verified to not be zero to prevent debt dust below the `minDebt` on the account that would have no value.

lekhovitsky:
> Such transaction could revert if the current USD conversion of the current `minDebt` value is equal to zero.

This scenario is a severe misconfig and is probably not really worth considering.

PR #262, opened 2024-08-18
fix: `ColleteralLogic.calcCollateral` no longer decrements token balances by `1`

lekhovitsky:
Fixes https://github.com/spearbit-audits/review-gearbox/issues/23

The fix only addresses the collateral calculation part.

As for withdrawals, keeping 1 unit of a token when a user passes default amount (`type(uint256).max`) is totally fine in terms of security implications while making future interactions with this credit account significantly cheaper. Sophisticated users are totally free to ignore this and pass the desired amount instead. Subtracting `1` in both branches of withdrawal actually makes things worse for those sophisticated users.


StErMi:
The new implementation has addressed the issue related to the `CollateralLogic.calcOneTokenCollateral` function.
The `CreditFacadeV3._withdrawCollateral` issue has been acknowledged by Gearbox and no changes have been applied.

PR #264, opened 2024-08-18
feat: make `botList` immutable in `CreditFacadeV3`

lekhovitsky:
Fixes https://github.com/spearbit-audits/review-gearbox/issues/72

### DIFF
--- a/contracts/credit/CreditConfiguratorV3.sol
+++ b/contracts/credit/CreditConfiguratorV3.sol
@@ -469,29 +469,12 @@ contract CreditConfiguratorV3 is ICreditConfiguratorV3, ACLNonReentrantTrait {
         emit SetPriceOracle(newPriceOracle); // I:[CC-21]
     }
 
-    /// @notice Sets the new bot list contract in the credit facade
-    /// @param newBotList New bot list
-    function setBotList(address newBotList)
-        external
-        override
-        configuratorOnly // I:[CC-2]
-    {
-        _setBotList(newBotList); // I:[CC-33]
-    }
-
-    /// @dev `setBotList` implementation
-    function _setBotList(address botList) internal {
-        CreditFacadeV3 cf = CreditFacadeV3(creditFacade());
-        if (botList == cf.botList()) return;
-        cf.setBotList(botList); // I:[CC-33]
-        emit SetBotList(botList); // I:[CC-33]
-    }
-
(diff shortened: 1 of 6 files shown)


StErMi:
@lekhovitsky 

1) should the `CreditConfiguratorV3.setCreditFacade` reverts if `prevCreditFacade.botList() != newCreditFacade.botList()`? Is there a specific reason to allow the new credit facade to have a different bot list (brand new and with a different configuration compared to the old one)? This would bring back all the side effects already listed in the https://github.com/spearbit-audits/review-gearbox/issues/72 finding.
2) Which was the reason to allow the `CreditConfigurator` to change the `botList` of an existing credit facade? Do you see any possible side effects of making it `immutable`?
3) Let's assume that there's a bug in the bot list used by the current credit facade, what would be your planned actions? Disable **every** bot configured, redeploy a new credit facade with a working bot list and call `CreditConfiguratorV3.setCreditFacade`?

Note that reverting when the new credit facade has a different bot list (like suggested in point 1) could be a problem if the current (old) credit facade has a faulty bot list that must be replaced by the bot list used of the new credit facade.

lekhovitsky:
@StErMi 

1. The original version of fixes reverted in case of mismatch but we decided to omit it eventually. We're well aware of consequences of replacing the bot list, but in case it contains some bug, or some crucial functionality upgrade is required, it would be nice to just update and audit a single contract instead of also having to upgrade and re-audit the credit configurator just to bypass this check.
2. The bots feature was "in making" and we were unsure that the original implementation of the bot list covered all our needs. For example, in v2.1, it contained [bot funding](https://github.com/Gearbox-protocol/core-v2/blob/d1ef1aa0b3ec94c832c01b81d30d706a4b7dc4c8/contracts/support/BotList.sol#L118), which we got rid of in v3.0. In v3.1 we've also dropped [bots with special permissions](https://github.com/Gearbox-protocol/core-v3/blob/ae093e204196127b57a950fbb0ba9bca85040693/contracts/core/BotListV3.sol#L90). Both changes broke the interface though, so we've never used its mutability anyways.
3. Yes. A more intricate question is communicating this to customers properly as replacing the bot list can invalidate outstanding stop-loss orders or deleverage services. I think this can be mitigated by making the new bot list inherit some non-broken configs of the old one. On the other hand, the bot list contract is extremely simple and can be formally verified to contain no bugs.

PR #266, opened 2024-08-18
fix: pool functions revert on no-ops and slippage

lekhovitsky:
Fixes https://github.com/spearbit-audits/review-gearbox/issues/36


StErMi:
The recommendations have been correctly implemented.

PR #269, opened 2024-08-18
fix: move LSB calculation inside `BitMask` library

lekhovitsky:
Fixes https://github.com/spearbit-audits/review-gearbox/issues/8


StErMi:
The PR correctly implements the recommendations of the finding.

PR #272, opened 2024-08-19
fix: facade functions revert on no-ops

lekhovitsky:
Fixes:
* https://github.com/spearbit-audits/review-gearbox/issues/96
* https://github.com/spearbit-audits/review-gearbox/issues/106

### DIFF
--- a/contracts/credit/CreditFacadeV3.sol
+++ b/contracts/credit/CreditFacadeV3.sol
@@ -387,7 +387,7 @@ contract CreditFacadeV3 is ICreditFacadeV3, ACLNonReentrantTrait {
         returns (uint256 seizedAmount)
     {
         address priceOracle = _priceOracle();
-        IPriceOracleV3(priceOracle).updatePrices(priceUpdates);
+        if (priceUpdates.length != 0) _updatePrices(priceOracle, priceUpdates);
 
         if (token == underlying) revert UnderlyingIsNotLiquidatableException();
         (CollateralDebtData memory cdd, bool isUnhealthy) = _revertIfNotLiquidatable(creditAccount);
@@ -528,8 +528,7 @@ contract CreditFacadeV3 is ICreditFacadeV3, ACLNonReentrantTrait {
                     // addCollateral
                     else if (method == ICreditFacadeV3Multicall.addCollateral.selector) {
                         _revertIfNoPermission(flags, ADD_COLLATERAL_PERMISSION); // U:[FA-21]
-                        (address token, uint256 amount) = abi.decode(mcall.callData[4:], (address, uint256)); // U:[FA-26A]
-                        _addCollateral(creditAccount, token, amount); // U:[FA-26A]
+                        _addCollateral(creditAccount, mcall.callData[4:]); // U:[FA-26A]
                     }
                     // addCollateralWithPermit
                     else if (method == ICreditFacadeV3Multicall.addCollateralWithPermit.selector) {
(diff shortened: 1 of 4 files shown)


StErMi:
@lekhovitsky the `partiallyLiquidateCreditAccount` function executes the [`_addCollateral`](https://github.com/Gearbox-protocol/core-v3/blob/spearbit-96-106/contracts/credit/CreditFacadeV3.sol#L971-L980) and [`_withdrawCollateral`](https://github.com/Gearbox-protocol/core-v3/blob/spearbit-96-106/contracts/credit/CreditFacadeV3.sol#L982-L991) functions that **do not** perform the check that would revert if `amount == 0`.

Is this behavior intended? This is the only case in the whole `CreditFacadeV3` that skips those checks.

Other than that, the code seems good and implements the recommendations from the findings.

lekhovitsky:
@StErMi If `amount` passed to `partiallyLiquidateCreditAccount` is `0`, the code would revert inside `_manageDebt`. This looks safer compared to reverting inside `_withdrawCollateral` if `feeAmount` rounds down to `0` for some reason.

contracts/credit/CreditFacadeV3.sol
```diff
@@ -729,17 +723,20 @@ contract CreditFacadeV3 is ICreditFacadeV3, ACLNonReentrantTrait {
...
-    function _withdrawCollateral(address creditAccount, address token, uint256 amount, address to) internal {
+    function _withdrawCollateral(address creditAccount, bytes calldata callData) internal {
+        (address token, uint256 amount, address to) = abi.decode(callData, (address, uint256, address)); // U:[FA-36]
+
         if (amount == type(uint256).max) {
             amount = IERC20(token).safeBalanceOf(creditAccount);
-            if (amount <= 1) return;
-            unchecked {
-                --amount;
+            if (amount >= 1) {
```
cryptarasecurity, line 731: Wouldn't that cause a revert if the balance of the credit account is `1`? What about previous gas optimisations used here?

lekhovitsky: I guess the point is to avoid no-op withdrawals when the balance is `<= 1` (one can still specify exactly `amount == 1` if this 1 wei is of any value).

StErMi: That's the main reason why I do not like to have custom logics like this one.

As @lekhovitsky said, people can just specify the **exact** number of tokens to withdraw to avoid this revert, and definitely it's an edge case (given that you have to have just `1 wei` in the balance of your CA for this token) but still...

@lekhovitsky with gas as low as `0.805 gwei` like we have right now, is it really worth having this custom behavior?

lekhovitsky: @StErMi Still don't see a single an issue with it, tbh

contracts/credit/CreditFacadeV3.sol
```diff
@@ -707,6 +700,7 @@ contract CreditFacadeV3 is ICreditFacadeV3, ACLNonReentrantTrait {
         uint256 forbiddenTokensMask
     ) internal returns (uint256, uint256) {
         (address token, int96 quotaChange, uint96 minQuota) = abi.decode(callData, (address, int96, uint96)); // U:[FA-34]
+        if (quotaChange == 0) revert AmountCantBeZeroException(); // U:[FA-34]
```
lekhovitsky, line 703: @StErMi probably for completeness this line should go after `quotaChange` is rounded; something like 

```solidity
quotaChange = quotaChange != type(int96).min
     ? quotaChange / int96(uint96(PERCENTAGE_FACTOR)) * int96(uint96(PERCENTAGE_FACTOR))
     : quotaChange;
if (quotaChange == 0) revert AmountCantBeZeroException(); // U:[FA-34]
```

StErMi: Yes, I think it makes sense

PR #280, opened 2024-08-22
feat: loss liquidator

lekhovitsky:
Lossy liquidations no longer pause credit facade. Instead, they must go through a contract that can enforce various policies on them. For more context, an early implementation can be found in the governance repo.

`liquidateCreditAccount` now returns reported loss.

Access modifiers of some functions are revised.


lekhovitsky:
Replicated by #285

PR #281, opened 2024-08-26
feat: phantom token withdrawals

lekhovitsky:
Notion of phantom tokens is added to the core.
When adding a phantom token as collateral, credit configurator checks that its deposited token is already added as collateral.
When withdrawing a phantom token in the credit facade, a call to a withdrawer adapter is made first to unwrap phantom token into its deposited token.

Number of optimizer runs is decreased to `1000` (which was actually always used in practice) for all contracts to fit into a size limit.


lekhovitsky:
Replicated by #286

PR #285, opened 2024-09-01
feat: loss liquidator

lekhovitsky:
Lossy liquidations no longer pause credit facade.
Instead, they must go through a contract that can enforce various policies on them. For more context, an early implementation can be found in the governance repo.

`liquidateCreditAccount` now returns reported loss.

Access modifiers of some functions are revised.

### DIFF
--- a/contracts/credit/CreditConfiguratorV3.sol
+++ b/contracts/credit/CreditConfiguratorV3.sol
@@ -506,8 +506,7 @@ contract CreditConfiguratorV3 is ICreditConfiguratorV3, ControlledTrait, SanityC
             (uint128 minDebt, uint128 maxDebt) = prevCreditFacade.debtLimits();
             _setLimits({minDebt: minDebt, maxDebt: maxDebt}); // I:[CC-22]
 
-            (, uint128 maxCumulativeLoss) = prevCreditFacade.lossParams();
-            _setMaxCumulativeLoss(maxCumulativeLoss); // I:[CC-22]
+            _setLossLiquidator(prevCreditFacade.lossLiquidator()); // I:[CC-22]
 
             _migrateEmergencyLiquidators(prevCreditFacade); // I:[CC-22C]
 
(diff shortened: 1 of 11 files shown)


cryptarasecurity:
# Notes

- Removed the `whenNotPaused` modifier for `lendCreditAccount` and `repayCreditAccount` functions on the `PoolV3` contract.
- The `whenNotPausedOrEmergency` modifier is checking for the `liquidateCreditAccount` and `partiallyLiquidateCreditAccount` can now be called by the `lossLiquidator` contract.
- The `liquidateCreditAccount` will no longer pause the contract if the cumulative loss from bad debt exceeds a threshold.
- The `liquidateCreditAccount` function can only be called by the `lossLiquidator` if the credit manager reports a different than zero `reportedLoss`.
- A new function named `setLossLiquidator` is created on the `CreditFacadeV3` and its corresponding `CreditConfiguratorV3` function `setLossLiquidator`, only callable by the configurator (`configuratorOnly`).
- The `resetCumulativeLoss` function was removed from the `CreditConfiguratorV3`  contract.
- The `removeEmergencyLiquidator` function on the configurator can now by called by the controller or the configurator (`controllerOrConfiguratorOnly`) instead of just the configurator.

# Issues

- When the `lossLiquidator` has not been set via the `setLossLiquidator` call, the `liquidateCreditAccount` function allows anyone to perform liquidations, even if a `reportedLoss` greater than zero is generated. In the previous implementation, the contract would automatically pause under such circumstances. However, since the pausing functionality has been removed, liquidations with losses can continue unchecked until the `lossLiquidator` is set or the contract is manually paused. Additionally, if the contract is paused and the `lossLiquidator` is still not assigned, emergency liquidators will retain the same privileges as the `lossLiquidator` to liquidate accounts during the emergency phase, regardless of a non-zero `reportedLoss`. This change introduces a risk of potential losses being realized without the proper oversight and failsafe mechanisms in place.

**Recommendation**: Implement a default restriction that prevents liquidations resulting in a non-zero `reportedLoss` unless the `lossLiquidator` is set. Specifically, require the `lossLiquidator` address to be non-zero before allowing liquidation calls that incur a loss. Additionally, in an "emergency situation" when the contract is paused, restrict liquidation permissions exclusively to the `lossLiquidator` to ensure tighter control over the process. An event should also be emitted to notify stakeholders whenever a liquidation occurs under these conditions, enhancing transparency and facilitating prompt corrective action.

PR #287, opened 2024-09-10
feat: extend contract types

lekhovitsky:
Having `contractType` return the full contract name makes more sense and doesn't cost anything since it shouldn't really be queried on-chain.

The general rule for naming is to use snake-cased capitalized name of the contract without the `V3` prefix.
The exceptions are contracts that implement interfaces `IAdapter`, `IPriceFeed`, `IBot`, `IRateKeeper`, `IInterestRateModel` (also `IZapper`, which is not included in this repo), which start with respective abbreviation.

### DIFF
--- a/contracts/core/AccountFactoryV3.sol
+++ b/contracts/core/AccountFactoryV3.sol
@@ -45,7 +45,7 @@ contract AccountFactoryV3 is IAccountFactoryV3, Ownable {
     uint256 public constant override version = 3_10;
 
     /// @notice Contract type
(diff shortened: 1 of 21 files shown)


lekhovitsky:
@Van0k guess we no longer need this one https://github.com/Gearbox-protocol/bots-v3/pull/7

cryptarasecurity:
# Notes

- `contractType` on all the contracts were extended to readable text instead of using ID's.

PR #289, opened 2024-09-13
feat: simplify `TumblerV3`'s deployment

lekhovitsky:
Previously, `TumblerV3`'s constructor was accepting `pool` as argument and reading `poolQuotaKeeper` from there. This caused complications in deployments since it required a configurator-only function `setPoolQuotaKeeper` to be called between contract creations. This PR changes the constructor to accept `poolQuotaKeeper` instead of the pool, which allows to deploy contracts first and configure them later.
Note that this is the only occurrence of such issue in core contracts.

### DIFF
--- a/contracts/pool/TumblerV3.sol
+++ b/contracts/pool/TumblerV3.sol
@@ -50,13 +50,15 @@ contract TumblerV3 is ITumblerV3, ControlledTrait, SanityCheckTrait {
     mapping(address => uint16) internal _rates;
 
     /// @notice Constructor
-    /// @param pool_ Pool whose quota rates to set by this contract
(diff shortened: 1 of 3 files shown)


cryptarasecurity:
# Notes

- This change allows the `setPoolQuotaKeeper` function in the pool to be called after deploying all the contracts. Previously, `TumblerV3` required a `PoolV3` contract, which had to have its `poolQuotaKeeper` set beforehand for the `TumblerV3` constructor to retrieve it. Now, `TumblerV3` directly accepts the `PoolQuotaKeeperV3` contract, removing the dependency on `PoolV3` to have the `poolQuotaKeeper` preconfigured.

PR #290, opened 2024-09-30
feat: fixes after secondary Spearbit review

lekhovitsky:


0xmikko (approved):

contracts/credit/CreditConfiguratorV3.sol
```diff
@@ -514,7 +514,8 @@ contract CreditConfiguratorV3 is ICreditConfiguratorV3, ControlledTrait, SanityC
             (uint128 minDebt, uint128 maxDebt) = prevCreditFacade.debtLimits();
             _setLimits({minDebt: minDebt, maxDebt: maxDebt}); // I:[CC-22]
 
-            _setLossLiquidator(prevCreditFacade.lossLiquidator()); // I:[CC-22]
+            address lossLiquidator = prevCreditFacade.lossLiquidator();
+            if (lossLiquidator != address(0)) _setLossLiquidator(lossLiquidator); // I:[CC-22]
```
Saw-mon-and-Natalie, line 518: Just a note that `migrateParams` needs to be false when using a newer version of `CC` with an old version of `CM` since the old `CF`s connected to old `CM`s don't have the `lossLiquidator()` endpoint.

So the params and some other configurations might need to be set manually.

